Issue #52, opened 2017-07-17
Unable to join Phillips Hue dimmer switch


jimduchek:
I'm attempting to get a Phillips Hue dimmer switch to talk to a HUSBZB-1.  Obviously a pretty untested setup, but I'd love to get this working, as there are just so few battery-powered switches/dimmers out there.   This is less an 'issue' and more just opening a discussion.  TLDR on this: It doesn't seem, out of the box, the switch is replying to discovery with it's own sender network ID, but with 0 as the sender.  So the stack doesn't see the replies.

Where we seem to be stuck is, we send out an IeeeAddressRequest:
02:51:29.704  DEBUG  TX CMD: IeeeAddressRequest [0/0 -> 2858/0, cluster=0001, TID=5C, nwkAddrOfInterest=2858, requestType=1, startIndex=0]

And we do get a response -- with a sender of 0, not 2858.  So it never 'sees' the response (as it expects the sender of the response to match the destination of the request).

02:51:29.740  TRACE  <-- RX ASH frame: AshFrameData [frmNum=0, ackNum=0, reTx=false, data=5C 90 45 00 00 00 01 80 00 00 40 01 00 00 2C FF 00 00 00 FF FF 0D 00 00 32 6F 19 02 01 88 17 00 2A 0B 00]
02:51:29.742  DEBUG  RX CMD: IeeeAddressResponse [0/0 -> 0/0, cluster=8001, TID=NULL, status=SUCCESS, ieeeAddrRemoteDev=0017880102196F32, nwkAddrRemoteDev=2858, numAssocDev=0, startIndex=null, nwkAddrAssocDevList=[]]
02:51:39.707  DEBUG  Ieee Address for 2858 returned null

I'm unclear as to why the sender isn't set -- I think it might be possible that the dongle may be intercepting the request (since it knows the 'right answer'?  AFAIK it should treat this as any other unicast command, but... it is not)  At any rate, I changed this in zdo/command/IeeeAddressRequest.java (I'm not sure this is a good idea for everybody, but for my setup, it doesn't matter for the time being, as it should be correct on anything but a router I think):

//        return (((IeeeAddressRequest) request).getDestinationAddress()
//                .equals(((IeeeAddressResponse) response).getSourceAddress()));
        return (((IeeeAddressRequest) request).getNwkAddrOfInterest()
                .equals(((IeeeAddressResponse) response).getNwkAddrRemoteDev()));

[...]

PR #53, opened 2017-07-17
Change ZDO response matcher to use NWKAddrOfInterest where available

cdjackson:
#52 

Requires further testing


cdjackson:
Merging as now tested.

PR #55, opened 2017-07-26
Autogenerate EzspFrame

cdjackson:


cdjackson:
This also fixes a couple of errors that came from the XML definition file resulting in some duplication of commands (eg callback and noCallback)

PR #74, opened 2017-10-07
Initial addition of server concept

cdjackson:
This adds a concept of servers to the framework. Servers are added to a node rather than the network to allow different servers per node. Eg for a time server, or IAS server, we may have a single network wide server instantiated, and it can be added to all applicable nodes. However for example, for OTA, a separate server per node is required.

Once servers are added, the node will handle the MatchDescriptor request to respond to any devices searching for servers.


cdjackson:
Will wrap this into the OTA server changes.

PR #80, opened 2017-10-13
Add initial definition of thermostat cluster

cdjackson:


cdjackson:
This will require some work to allow multiple schedules to be set.

PR #87, opened 2017-10-16
Fix handling of ZclHeader (de)serialisation

cdjackson:
This fixes a bug in the processing of the ZclHeader fields which incorrectly handles some fields.


cdjackson:
Reverted this PR to just fix the bug in the ZclHeader processing. Unfortunately the direction processing needs a bit more work as it is the wrong way around once we start to need to interact with remote clients rather than servers. This will be added in a separate PR.

PR #88, opened 2017-10-23
Command listener implementation 

saukijan:
Implemented a command listener on ZclCluster
implemented ZCL_Status(OK)


cdjackson:
Hi,
I'm not sure how best to handle this PR. I think at least it needs tidying as there seem to be a number of files that are showing differences, but I think they have just been reformatted (ie all the pom files for starters). This is causing conflicts in these files at least.

Since I don't think you've actually changed much (based on our discussion on Sunday) it would be good if you could update the PR so that the only changes showing in the diff are the ones relating to the PR (ie the addition of the command listener). That should remove most of the conflicts and the clean things up.

The other possibility (maybe the easier option) is that I could add the command listener code into my latest changes. I think this is only a "few lines" of code to add the listener to the command and the other change to add the ZclStatus in the desrialiser is something that is already added.

Let me know what you think is best/easiest ;)

Cheers
Chris

saukijan:
Hi Chris,

Yes, it's only a few lines of codes on the library side, I haven't had the
time to clean up the binding side so I didn't do a pull request for it yet.

It would probably be easiest to just change those few lines on your side, I
had some problems setting up the zsmartsystems project on my machine, so
that might have caused some problems when I did a push to my fork on
github.

I'll try looking over the ZigBee binding as soon as possible, hopefully,
I'll get it done by the end of this week.

Dovydas

PR #93, opened 2017-11-02
Serial port extensions to support event driven communications

cdjackson:
This fundamentally adds two methods (```read``` and ```write```) to the ```ZigBeePort``` interface to remove the need to use ```InputStream``` and ```OutputStream``` which do not have non-blocking options which makes it difficult to close the port.

The reference implementation of the ```ZigBeeSerialPort``` is written to use serial event listener for receive which allows a thread to wait for data, and be woken again. A version of ```read(timeout)``` is also provided to allow the method to return if no data is received within ```timeout``` milliseconds.

This has been tested against the Telegesis dongle so far.

@cschwer I'd welcome any comments since I noticed you were changing the serial port usage anyway.


cschwer:
I will test it out with ember.  Yes I was changing the serial port implementation because I was having the same trouble with closing the port.

cdjackson:
Thanks. I've actually just done a quick test with all dongles and it seems to work, but I'd appreciate any further comments... I'd like to merge this tomorrow to allow it to feed in to some other enhancements.

cdjackson:
I've just done a few more checks with the Telegesis dongle and made sure that the handlers close ok. If there are issues with the other dongle handlers, then it would be worth checking the Telegesis implementation.

Any further issues can be dealt with as subsequent PRs.

PR #94, opened 2017-11-03
WIP: Implement OTA firmware upgrade cluster and server

cdjackson:
This implements the OTA cluster, and an OTA server. It also modifies the way that the direction bit is handled, as well as the handling of input and output clusters in the endpoint.


cdjackson:
This has been tested as working - an OTA file has been transferred. I'll merge now to allow other updates to be made and any further refinement of the OTA interfaces that may be required will be added as separate PRs. It's possible that there will be further changes needed once this is integrated into the application.

PR #97, opened 2017-11-07
close ashHanlder before closing serial port.

cschwer:
closing serialPort but leaving ashHandler running can lead to an exception if using the rxtx library instead of jssc for serial port interface.  This PR closes the ashHandler before attempting to close the serial port.


cdjackson:
Do you think ```setClosing()``` should be called first?  I looked at this code for the Telegesis dongle the other day and just checked your implementation and it throws an error.

The following seems the cleanest for me here -:

```
    public void shutdown() {
        ashHandler.setClosing();
        serialPort.close();
        ashHandler.close();
    }
```

* ```setClosing``` tells the handler to exit.
* ```serialPort.close()``` closes the port and should exit any waiting read request.
* ```close``` then can free everything up and wait for the threads to terminate.

I use RxTx in my app and don't have any problems with this in Telegesis at least.

What do you think?

cschwer:
Thats fine with me. I just tested it out with rxtx as well and it works.

cdjackson:
Thanks.

PR #110, opened 2017-11-13
Updates to fix issues with sending ezsp frames.

cschwer:
Added check that timerTask is running.
add reset for retries counter
drop message if retry count is exceeded
wrap sendRetry in try/catch so that retry timer does not get killed.

Signed-off-by: Charles Schwer charles.schwer@gmail.com


cdjackson:
I've added a CLA to the project - just to protect everyone so say that any contributions can be  used without copywite issues. It should be reasonably simple to click on the button that you hopefully see (assuming you agree of course ;) ).

cschwer:
Yep, I saw the CLA and figured as much.  I have signed it already, hopefully I did that right.

cdjackson:
I think this looks fine and clearly you're using the Ember interface more than me at the moment. Can I just get you to add a signed-off-by signature to the PR please.

cschwer:
yep, added signature to original comment.

PR #116, opened 2017-11-16
Configuration update concept

cdjackson:
This adds

```
    void updateTransportConfig(TransportConfig configuration);
```

to the ```ZigBeeTransportTransport``` interface to allow a flexible interface for passing configuration to the transport layer.

Trust centre configuration for the Telegesis dongle is also included in this PR.

@cschwer what do you think - not tested properly yet - I just refactored the trust centre join mode code to this, but comments welcome...


cschwer:
so far so good.. once you have it merged I can add some of the other configuration options for ember.
:+1:

cdjackson:
Ok, I'll try and get this merged later today. My initial thought is to keep this interface for general configuration (ie non-dongle specific) but let's see. The problem if we allow dongle specific stuff in here is that dongle drivers written outside of this repo would need to use another interface anyway so it's a little messy...

cschwer:
ok, I think I'm following you on this.   So there are several configuration items in ZigbeeDongleEzsp, 
ie: 
EZSP_CONFIG_TX_POWER_MODE, 
EZSP_CONFIG_MAX_HOPS, 
EZSP_CONFIG_MAX_END_DEVICE_CHILDREN, 
etc

Would those be configured via this method, and if not is there some other way to make those values be configurable in the library?  Right now they are hard coded to values.

cdjackson:
Yeah - these are pretty generic so we can make them configurable through this with some generic names.

Issue #109, opened 2017-11-13
Support EZSP version 5 and later.


haojiesong:

cdjackson:
I believe that EZSP5 is simply an extension of EZSP4 with a number of new commands, so this should already be supported in general. If there are new commands that you require implementing, then please list them and I'm sure they can be added.

If there's something more general that has changed then I'll take a look, but I think it's just an extension of 4?

haojiesong:
From EmberZNet 5.9.x, the EZSP version has upgraded from 4 to 5. It is incompatible and need firstly do special handling.

cdjackson:
Ok, from what I could see there was minimal difference - UG100 doesn't seem to indicate anything special - can you provide a reference to the differences as I can't seem to find anything from a quick search.
Thanks.

cdjackson:
Ok, I see in the protocol format (section 3) of UG100 that there is now an extra byte in the protocol. Strangely this "small" change doesn't feature in the documents "what's changed" list!!!

I'll take a look at adding support for this in the near future as I might flash one of my devices with the latest NCP firmware.

PR #122, opened 2017-11-19
WIP: Ember ASH v3 EZSP v5

cdjackson:
This splits the frame formatting out of the ASH frame, and the packet handler out of the FrameHandler and puts it into a separate file. So we have AshFrameHandler as an abstract class, and AshFrameHandlerV2 to cover the frame and packet specific to this version. A separate AshFrameHandlerV3 is also started to handle the new version.

I hadn't intended on doing this right now, but I found out the hard way that Silabs have made huge changes as I upgraded to NCP6 while testing the dongle firmware update code.

@cschwer you might want to have a look at this - any comments welcome. I won't merge this for a while unless it interacts with the dongle update code while I will need to get merged quite soon.

Partly addresses #109...


cdjackson:
Tested EZSP5 with NCP 5.10.1

cschwer:
I've tested this with ash v2 successfully.  I noticed that the AshFrameHandlerV3 class is not used in the code, only in the tests.  Were you planning on instantiating it if the EZSP version request comes back as v5?

cdjackson:
Cool - thanks.

I think there will need to be a constructor argument (or something) to define ASHV2 or V3. V3 is quite different and I think when I loaded V3 firmware it didn’t respond at all to the V2 reset / version request so I don’t think it’s something that can be detected automagically.  I’ll look at this more when I take another look at V3...

cdjackson:
> Were you planning on instantiating it if the EZSP version request comes back as v5?

I should have also added that ASH V2/V3 and EZSP V4/V5 are not linked. Ember 5.8 is ASH V2/EZSP V4, 5.9 is ASH V2/EZSP V5, and 6.0 is ASH V3/EZSP V5.........

sebastienmarion:
Hi,
I want to correct ezsp v5 and v6 continue to use ash v2.
ASH v3 is for thread NCP stack not for zigbee stack.
I use stack v6.0 : ie ezsp v6 and ashv2, I have add patch on my fork to manage from ezsp v4 to v6 currently. 
I have not manage secure connexion on ezsp v5/v6, maybe later, I focus on split initialization to be able to use zigbee ezsp dongle in another context.

cdjackson:
> ASH v3 is for thread NCP stack not for zigbee stack.

Where have you seen this? From my experience running NCP6.0.0 it required ASHv3.

This PR will likely be refactored at some stage, but there should be nothing to prevent running ASH V2.

sebastienmarion:
I found this information in document UG101 rev 0.9 provide with stack zigbee_6.1
In the same gecko sdk suite v2.1 I found in stack thread v2.5 a document UG115 rev0.4 who describe ashv3 protocol
My current ezsp dongle is built with stack zigbee 6.0 and work correctly with ashv2, I only add support for extended header in ezsp protocol.

cdjackson:
I don't recall this being the case when I tested it, but even if it is, I'm not sure what you're actually reporting with your comment? ASHV2 should work fine already.

PR #181, opened 2018-01-14
Add endpoint before initialising network

cdjackson:
The ember endpoint must be added before the network is initialised. The fact that this was failing probably means it's not required, but for now lets leave it here.


cschwer:
Beat me to it, I was going to submit a PR for this today!  :+1:

cdjackson:
Thanks @cschwer. As a matter of interest how are you finding the stability of the Ember driver? openHAB seems to have some issues - possibly related to TX queue. As you've been using the Ember more than me recently if you have any thoughts, it would be welcome ;)

cschwer:
@cdjackson It has been fairly stable, the issues i'm seeing are:

I'm still having issues with startup, the occasional EMBER_NOT_JOINED message. 

Recently I've found that if i send a management permit joining request and then send a EzspGetEui64Request, that second request gets stuck sending over and over.  I haven't had a time to track down exactly why that is.

Also I've seen if there is a high load on the machine, the ember will send back an error 51:  "Error: Exceeded maximum ACK timeout count"   and the stack will need to be reset by sending an RST frame.  I have the RST being sent back on error,, I still need to report that back up the chain so the whole stack can be reinitialized.

PR #201, opened 2018-02-01
Extend ezsp frame

sebastienmarion:
Add new ezsp frame and configuration value.


cdjackson:
I will likely reject this as these classes will be automatically coded. I've been adding additional classes recently although they haven't been added to a PR yet - I'll try and do this in the coming days and if there are any more missing that are required we can look to add them.

sebastienmarion:
Ok for automatically coded but where is the generator ?
out of subject : I am very new in github world and open source project, so how can I contribute ?

cdjackson:
The coder is not part of the project at the moment (for a number of reasons). I’ll likely be able to add it at some stage, but want to avoid that right now. I will push all the additional classes I’ve currently added shortly (or maybe tonight as I’m about to get on a flight back to the UK from Paris…).

I’d suggest to raise issues if you want to add something - it avoids multiple people working on the same issue. Also, if possible keep the PRs small - eg one feature update per PR if possible (I know sometimes it’s not possible as there are dependancies - it just makes it easier to review).

I’ll try and add this to the contribution guidelines.

(oh, and contributions would be great - thanks :) ).

sebastienmarion:
I discover this so I have done branch from my master to split in two functionalities. I can split more.
I have add essentially support of MfgLib that is usefull for doing a 802.15.4 sniffer.
I have add missing constant for stack configuration.

cdjackson:
Closed by #202 

If there are still missing commands, then please open an issue and I'll add them to the definition. I think GP commands aren't in the list yet so you'll be wanting these at least :) .

PR #202, opened 2018-02-01
Add additional Ember command classes

cdjackson:


cdjackson:
FYI @sebastienmarion

sebastienmarion:
A question, since this command is generate by tool, what is the good way to add to function to use it simpler ?
example for EzspMfglibSendPacketRequest, this frame is compose by a length, a table and a crc.
a useful function is to pass only the message to send and internally it compute length and create a bigger table for include crc byte.

cdjackson:
> example for EzspMfglibSendPacketRequest, this frame is compose by a length, a table and a crc.
a useful function is to pass only the message to send and internally it compute length and create a bigger table for include crc byte.

The tool can generate this - if it's not generated it in this way at the moment then it's just because I haven't configured it properly. I'll take a look at this command and add the construct to generate the auto sizing.

PR #246, opened 2018-03-29
Don't process XBee responses that have a status that is not OK

cdjackson:
If an XBee response has an error, then don't process the remainder of the frame.


cdjackson:
Resolves the exception in #247

PR #274, opened 2018-04-09
Implement the Ember SPI protocol

### LINKED ISSUE #199: Ember cannot be initialized
For our project we're using com.zsmartsystems.zigbee in combination with org.openhab.binding.zigbee and trying to create a bridge for the Ember-Controller. The bridge itself is created but it's never set in status ONLINE. We stepped with a debugger through the code and the summary of our investigation is that the ZigBee initialization apparently gets stuck because the AshFrameHandler never receives a com.zsmartsystems.zigbee.dongle.ember.ash.AshFrame.FrameType.RSTACK an therefore can't be completed. So the bridge stays in status Status=INITIALIZING.

Below you see the key points which describe how the ZigBee initialization gets stuck and does therefore not complete. The description starts in com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp.initialize(). Since the process is thread-driven some things can happen in a slightly different order or simultaneously. But finally it ends up in the same situation.

In com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp.initialize()
- The statement if (!serialPort.open()) returns true.
- The statement ashHandler.start(serialPort) starts the thread to read packets from a stream. Initially there is no data available and the thread goes into wait state at org.openhab.binding.zigbee.internal.ZigBeeSerialPort.read(int)
- The statement ashHandler.connect() sends a reset frame. See below what happens in com.zsmartsystems.zigbee.dongle.ember.ash.AshFrameHandler.outputFrame(AshFrame).

In com.zsmartsystems.zigbee.dongle.ember.ash.AshFrameHandler.outputFrame(AshFrame)
- the for-loop
  - com.zsmartsystems.zigbee.dongle.ember.ash.AshFrameRst.getOutputBuffer() returns [26, 192, 56, 188, 126]. So these values represent the reset frame, right? Is 126 <=> ASH_FLAG_BYTE (end of a frame)?
  - port.write writes the content of the buffer byte by byte into the output stream gnu.io.RXTXPort$SerialOutputStream
  - the invocation of port.write leads to at least one invocation of org.openhab.binding.zigbee.internal.ZigBeeSerialPort.serialEvent(SerialPortEvent) depending on whether it is already in execution. It can happen up to five times since port.write is invoked five times.
 
In org.openhab.binding.zigbee.internal.ZigBeeSerialPort.serialEvent(SerialPortEvent)
- When serialEvent(SerialPortEvent) is called gnu.io.RXTXPort$SerialInputStream.read() delivers the values 3, 0, 167 five times in total.
- All the occurrences of 3, 0, 167 are stored in a buffer of type int[]. This results in buffer = [3, 0, 167, 3, 0, 167, 3, 0, 167, 3, 0, 167, 3, 0, 167, 0...]
- When all the available data is processed this.notify() is invoked which wakes up the thread that is waiting in org.openhab.binding.zigbee.internal.ZigBeeSerialPort.read(int). Each a byte will be read out there and returned to com.zsmartsystems.zigbee.dongle.ember.ash.AshFrameHandler.getPacket()

In com.zsmartsystems.zigbee.dongle.ember.ash.AshFrameHandler.getPacket()
- port.read() returns a byte from the buffer populated in org.openhab.binding.zigbee.internal.ZigBeeSerialPort.serialEvent(SerialPortEvent) and waits for ASH_FLAG_BYTE in order to return the collected data.
- Since buffer doesn't contain ASH_FLAG_BYTE the thread gets stuck in the loop and will never return to the caller.
- So the caller (thread "AshFrameHandler") will never receive com.zsmartsystems.zigbee.dongle.ember.ash.AshFrame.FrameType.RSTACK and consequently not proceed with the initialization process and set "stateConnected" to true. So no further communication is possible since it depends on this variable being set to true.

cdjackson:
This implements the Ember SPI protocol. It currently polls the callbacks every 250ms and will continue to poll immediately if further callbacks are outstanding.

Fixes #199 - @TomTravaglino you might like to take a look. I've currently not tested it extensively other than to make sure the network comes up and communications seems to work ok. I did get an ST motion sensor to join, so I think it's good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The library supports the Texas Instruments ZNP protocol over a serial interface.
 
 ### Silicon Labs Ember EM35x / EFR32
 
-The library supports the Silicon Labs EZSP protocol using ASH over a serial interface.
+The library supports the Silicon Labs EZSP protocol using ASH or SPI protocols over a serial interface. The implementation of the SPI protocol assumes that the SPI provides a TTY-like software interface to the application, or is otherwise abstracted via the ```ZigBeePort``` interface.  
 
 It is worth noting that EM3588 devices that have an embedded USB core will likely work with any baud rate, where dongles using external USB interface (eg CP2102 used with an EM3581) will likely require a specific baud rate. This has been noted on the HUSBZB-1 which embeds an EM3581 and requires a rate of 57600.
 
--- a/com.zsmartsystems.zigbee.console.ember/src/main/java/com/zsmartsystems/zigbee/console/ember/EmberConsoleNcpVersionCommand.java
+++ b/com.zsmartsystems.zigbee.console.ember/src/main/java/com/zsmartsystems/zigbee/console/ember/EmberConsoleNcpVersionCommand.java
@@ -27,7 +27,7 @@ public class EmberConsoleNcpVersionCommand extends EmberConsoleAbstractCommand {
 
     @Override
     public String getDescription() {
-        return "Gets the NCP version";
+        return "Gets the NCP firmware version";
     }
 
     @Override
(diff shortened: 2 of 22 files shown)


tomhoefer:
Thank you very much for this PR. 

@TomTravaglino pls let us discuss the corresponding test of the SPI integration

cdjackson:
Rebased to include all the latest Ember commands.

PR #288, opened 2018-04-20
Fix main class reference in console main manifest

cschwer:
Fixes error when running console main jar from command line.
"Error: Could not find or load main class com.zsmartsystems.zigbee.console.ZigBeeConsoleMain"

Signed-off-by: Charles Schwer charles.schwer@gmail.com


cdjackson:
Thanks.

Issue #289, opened 2018-04-24
Maven build fails when run in submodules


hsudbrock:
The maven build fails when run in a submodule (e.g., when running `mvn clean install` in the subfolder com.zsmartsystems.zigbee.console). The problem is that the `license-maven-plugin` cannot read the file `/src/main/header.txt`, which is defined in the parent module.

I don't think that this is a severe issue, but it can be a little annoying when one has to build all modules all the time.

There are several ways to resolve this (cf. https://github.com/mycila/license-maven-plugin/issues/36#issuecomment-33956023).

My suggestion is to disable the `license-maven-plugin` in the submodules as suggested in the discussion linked above, i.e., by setting the `inherited` property of the plugin configuration to `false`, and setting its configuration property `aggregate` to `true` (so that the plugin will check the source files in all submodules). I will add this small change as a PR, so one can see this written down.

PR #290, opened 2018-04-24
Disable the license-maven-plugin in submodules

hsudbrock:
This PR implements one solution to resolve https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/289, by:
* The license-maven-plugin is now disabled in submodules.
* When run in the parent module, the plugin will now also check source files in all submodules.

Another option to resolve the issue would be to reference the required license file explicitly from each submodule.

What do you think about using this solution?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,8 +138,10 @@
 				<groupId>com.mycila</groupId>
 				<artifactId>license-maven-plugin</artifactId>
 				<version>3.0</version>
+				<inherited>false</inherited>
 				<configuration>
 					<header>src/etc/header.txt</header>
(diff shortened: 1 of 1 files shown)


hsudbrock:
With this change, the license check is only executed when maven is executed in the top-level folder of the com.zsmartsystems.zigbee project, The check would no longer be executed when maven is executed in a subdirectory like, e.g., com.zsmartsystems.zigbee.dongle.xbee or com.zsmartsystems.zigbee.console.

cdjackson:
Ok, sounds good. Thanks.

PR #291, opened 2018-04-24
Fix main class name in console.main module

hsudbrock:
The package of the ZigBeeConsoleMain class was incorrect in the mainClass field for the console app jar file (probably leftover from module split), this is fixed here.


hsudbrock:
Then I was too fast (with this PR) and too slow (with updating my git repository)...
I will close this PR.

Issue #296, opened 2018-04-27
License goal not working


cdjackson:
@hsudbrock the change you made to the pom seems to complete stop the license formatter... 

Any thoughts on how to resolve this so it works in both scenarios?

hsudbrock:
I just had a deeper look into the mycila license-maven-plugin; the problem here appears to be the interplay between its `aggregate` setting (which I introduced with my change in the pom), and its `includes` setting.

Context: According to the plugin doc for the `aggregate` flag: "You can set this flag to true if you want to check the headers for all modules of your project. Only used for multi-modules projects, to check for example the header licenses from the parent module for all sub modules."

However, the `includes` setting appears to require path patterns relative to the parent pom, not relative to the submodule poms (which differs from the setting with `aggregate` being set to false, where the inherited configuration will interpret the path patterns in the `include` property relative to the submodule poms).

So what will solve this bug is to specify the `include` pattern relative to the parent pom, I'll add a PR for that.

With this change, `mvn com.mycila:license-maven-plugin:format` and `com.mycila:license-maven-plugin:check` should both work as expected (tested on my machine).

cdjackson:
Closed by #297

PR #297, opened 2018-04-27
[296] Fixes the include setting for the license goal

hsudbrock:
In https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/290 the mycila maven-license-plugin was configured so that it will consider also files in submodules when being executed in the parent module, and will not run in the submodules anymore.
However, the includes property was not appropriately changed as well. The includes property needs to specify the source files relative to the parent module now (while it had to specify them relative to the submodules before).

This commit adapts this by prepending a "*/" to the includes specification.

Bug: https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/296

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,7 @@
 					<failIfMissing>true</failIfMissing>
 					<strictCheck>true</strictCheck>
 					<includes>
-						<include>src/**/*.java</include>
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks @hsudbrock

PR #325, opened 2018-06-16
Minor refactoring of transaction timeout handling

cdjackson:
This refactors the management of transaction timeouts as discussed [here](https://github.com/openhab/org.openhab.binding.zigbee/issues/78). This should avoid possible lockouts by removing the multiple thread syncing on the same object.

It moves transaction classes into a new package and uses a new ```ZigBeeTransaction``` class to manage the transaction rather than using the embedded methods in the```ZigBeeTransactionManager``` which will make the transaction refactoring a little easier when this is finally done.

@rbi this is similar to your suggestion - any comments or testing welcome.


rbi:
I looked at your branch. This should fix all dead-lock issues that I tried to fix in my branch too.

In my branch I also modified `ZigBeeCommandNotifier` to simplify `addCommandListener` and `removeCommandListener`. That should fix another race condition. As `notifyCommandListeners` is `synchronized` I guessed that the whole  `ZigBeeCommandNotifier` class must be thread safe. I thought that's why the lists where copied in `addCommandListener`. With that implementation it is however possible to loose command listeners in the following scenario:

* Two threads add a new command listener
* Both create a copy of the current `commandListeners` list and both add a new command listener to their own copy
* Thread 1 sets the `commandListeners` field to its own copy with its new listener
* Thread 2 sets the `commandListeners` field to its own copy with its new listener
* The new command listener of thread 1 is lost.

Instead of just synchronizing the whole methods it would maybe also be possible to use one of the `Concurrent*` collections (`ConcurrentSkipListSet`, `ConcurrentLinkedQueue` or maybe even `CopyOnWriteArrayList`) but I think just synchronizing the methods works fine too.

cdjackson:
Well spotted - I’m not sure why it was done like this (I think it came from previous code) but I agree with your analyses and I’ll synchronise these methods.

rbi:
I think I managed to install a self-compiled version of this branch into my openhab. Uninstalling `com.zsmartsystems.zigbee` 1.0.11 and putting `com.zsmartsystems.zigbee` 1.0.13- into the addons folder so that it is the only active version should suffice or should it? If I done it right I can report if it works.

cdjackson:
Yes - that should be fine. So long as it’s the only active version, then it should be the right one. I’m also using it here and so far it’s ok.

PR #329, opened 2018-06-21
Refactor ember packages to export EZSP commands and structures

### LINKED ISSUE #328: Ember console uses classes from internal Ember dongle packages
Classes from the module `com.zsmartsystems.zigbee.console.ember` use classes from internal packages from the module `com.zsmartsystems.zigbee.dongle.ember` (viz. the package `com.zsmartsystems.zigbee.dongle.ember.internal.ezsp` and sub-packages).

This leads to problems (in particular in the context of this [issue](https://github.com/openhab/org.openhab.binding.zigbee/issues/219)) when using these modules as OSGi bundles, because the internal packages are not declared as exported in the manifest file that is generated by the maven build (as the maven build adds export statements for all bundles, unless bundles that contain `internal` or `impl` - cf. the [maven plugin docs](http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html)).

The simplest approach for solving this would be to simply configure the maven build to make the generated manifest export all bundles for the ember dongle module. This could by done by adding the following to the pom file of that module:
```
<build>
   <plugins>
    <plugin>
        <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
         <configuration>
           <instructions>
             <Export-Package>com.zsmartsystems.zigbee.dongle.ember.*</Export-Package>
           </instructions>
         </configuration>
    </plugin>
   </plugins>
 </build>
```

Exporting only packages without `internal`/`impl` except the affected packages appears to be more difficult to me.

Another option would be to rewrite the ember console / the ember dongle modules such that imports of internal classes are no longer needed - this option also seems not so easy.

cdjackson:
@hsudbrock please take a look at this - fundamentally it moves all the Ember structures and commands out of the internal package so I think should fix your issue.

Fixes #328 


hsudbrock:
Thanks @cdjackson - yes, this solves my issue, and looks like a sensible solution.

PR #330, opened 2018-06-22
Add module to maven build that creates a p2 repository with built bundles

hsudbrock:
The p2 repository will be created in folder releng/p2repo/target/repository if the maven profile 'p2repo' is active (e.g., when running `mvn clean install -Pp2repo`).

This p2 repository is, imho, quite useful when building OSGi applications like the openhab ZigBee binding with maven, in case one wants to build the openhab ZigBee binding against a self-compiled version of this zigbee core library

If you think that this is useful functionality to be included in the zsmartsystems zigbee library, I'd be happy if this PR gets merged.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,16 @@
 		<module>com.zsmartsystems.zigbee.serial</module>
 		<module>com.zsmartsystems.zigbee.test</module>
 	</modules>
+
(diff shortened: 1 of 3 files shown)


cdjackson:
Thanks @hsudbrock

PR #347, opened 2018-08-02
Move debug log message for unsuccessful leave command to correct place.

hsudbrock:
When trying to make a non-reachable node from the network I did not get any debug-level logging. I believe that the log message that is already in the code needs to be moved to the enclosing `if` statement, and did this in this PR.

Correct?


cdjackson:
Thanks Henning. This looks good - can you run the formatter on it please. It looks like there are 2 spaces in the debug message you added, rather than 4?

hsudbrock:
Yes, of course - I did not have this on my radar yet... Will update the PR then.

hsudbrock:
Applied the formatter (it was a tab instead of 4 spaces...).

cdjackson:
Thanks.

PR #363, opened 2018-08-13
Remove nodes from set of discovered nodes when they are removed from the network.

### LINKED ISSUE #362: Removed nodes are not removed from list of discovered nodes
The class `ZigBeeNetworkManager` puts all nodes for which discovery has completed into the set `nodeDiscoveryComplete`. This set is used to select between doing a `nodeAdded` and a `nodeUpdated` callback in the `updateNode` method.

When a node is removed, it is not removed from the `nodeDiscoveryComplete` set, resulting in a different callback behavior when a node is added for the first time (where `nodeAdded` is called twice, once when the node is added, and once when node discovery is complete), and when a node is added again after it has been removed before (where `nodeAdded` is called once, and after complete discovery `nodeUpdated` is called).

hsudbrock:
Resolves #362 by removing nodes from the set `nodeDiscoveryComplete` when they are removed from the set of network nodes.


cdjackson:
LGTM - thanks @hsudbrock

PR #370, opened 2018-08-16
Add a console command for displaying supported commands for a given cluster

hsudbrock:
Provides a new console command `cmdsupported` for showing commands supported by a cluster (analogously to the console command `attsupported` that does this for supported attributes).

The command can show only the supported generated commands, only the supported received commands, or both. 

Basically, the command calls the `ZclCluster#discoverCommands(Generated|Received)` methods to
trigger discovering the information from the device, and then the `ZclCluster#getSupportedCommands(Generated|Received)` methods for reading the result of this discovery.

### DIFF
--- a/com.zsmartsystems.zigbee.console.main/src/main/java/com/zsmartsystems/zigbee/console/main/ZigBeeConsole.java
+++ b/com.zsmartsystems.zigbee.console.main/src/main/java/com/zsmartsystems/zigbee/console/main/ZigBeeConsole.java
@@ -23,7 +23,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
-import java.util.concurrent.Future;
 
 import org.apache.commons.lang.StringUtils;
 
@@ -126,7 +125,7 @@ public final class ZigBeeConsole {
     /**
      * Constructor which configures ZigBee API and constructs commands.
      *
-     * @param dongle the dongle
+     * @param dongle            the dongle
      * @param transportCommands
      */
     public ZigBeeConsole(final ZigBeeNetworkManager networkManager, final ZigBeeTransportTransmit dongle,
@@ -175,8 +174,6 @@ public final class ZigBeeConsole {
         commands.put("rediscover", new RediscoverCommand());
 
         commands.put("stress", new StressCommand());
-
-        commands.put("cmdsupported", new CmdSupportedCommand());
 
         newCommands.put("nodes", new ZigBeeConsoleNodeListCommand());
         newCommands.put("endpoint", new ZigBeeConsoleDescribeEndpointCommand());
@@ -295,7 +292,7 @@ public final class ZigBeeConsole {
      * Processes text input line.
      *
      * @param inputLine the input line
-     * @param out the output stream
+     * @param out       the output stream
      */
     public void processInputLine(final String inputLine, final PrintStream out) {
         if (inputLine.length() == 0) {
@@ -309,7 +306,7 @@ public final class ZigBeeConsole {
      * Processes input arguments.
      *
      * @param args the input arguments
-     * @param out the output stream
+     * @param out  the output stream
      */
     public void processArgs(final String[] args, final PrintStream out) {
         try {
@@ -332,8 +329,8 @@ public final class ZigBeeConsole {
      * Executes command.
      *
      * @param networkManager the {@link ZigBeeNetworkManager}
-     * @param args the arguments including the command
-     * @param out the output stream
+     * @param args           the arguments including the command
+     * @param out            the output stream
      */
     private void executeCommand(final ZigBeeNetworkManager networkManager, final String[] args, final PrintStream out) {
         final ConsoleCommand consoleCommand = commands.get(args[0].toLowerCase());
@@ -394,7 +391,7 @@ public final class ZigBeeConsole {
     /**
      * Gets destination by device identifier or group ID.
      *
-     * @param zigbeeApi the ZigBee API
+     * @param zigbeeApi             the ZigBee API
      * @param destinationIdentifier the device identifier or group ID
      * @return the device
      */
@@ -422,7 +419,7 @@ public final class ZigBeeConsole {
     /**
      * Gets device by device identifier.
      *
-     * @param zigbeeApi the ZigBee API
+     * @param zigbeeApi        the ZigBee API
      * @param deviceIdentifier the device identifier
      * @return the device
      */
@@ -459,8 +456,8 @@ public final class ZigBeeConsole {
          * Processes console command.
          *
          * @param zigbeeApi the ZigBee API
-         * @param args the command arguments
-         * @param out the output PrintStream
+         * @param args      the command arguments
+         * @param out       the output PrintStream
          * @return true if command syntax was correct.
          */
         boolean process(final ZigBeeApi zigbeeApi, final String[] args, PrintStream out) throws Exception;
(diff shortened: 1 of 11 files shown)


com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleCommandsSupportedCommand.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public void process(ZigBeeNetworkManager networkManager, String[] args, PrintStream out)
+            throws IllegalArgumentException, InterruptedException, ExecutionException {
+        if (args.length < 3 || args.length > 4) {
+            throw new IllegalArgumentException("Invalid number of arguments");
+        }
+
+        final ZigBeeEndpoint endpoint = getEndpoint(networkManager, args[1]);
+        final Integer clusterId = parseCluster(args[2]);
+        ZclCluster cluster = getCluster(endpoint, clusterId);
```
cdjackson: I'm not sure this is a good way to do this here. I know I've done similar things in other commands, but recently I've started to move away from just checking both input and output clusters (as is done in the ```getCluster``` method here.

The problem is that the method checks input clusters, then if there is no input cluster, it will check the output clusters. However, it means that you can't get the commands for an output cluster, if the input cluster exists.

In some recent commands, I've added a first parameter which is IN|OUT to allow this to be specified. Maybe we need to come up with a base method to do this consistently.

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleCommandsSupportedCommand.java
```diff
@@ -0,0 +1,111 @@
...
+
+        if (cluster == null) {
+            out.println("Cluster not found");
```
cdjackson: I thought that in general I was throwing an exception if there was an error? I guess you copied this from the attribute command, so presumably it's not consistent everywhere. What do you think? My original thinking was always to throw exceptions with errors to allow the implementation to catch these given there is no return.

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleCommandsSupportedCommand.java
```diff
@@ -0,0 +1,111 @@
...
+        if (result == null) {
+            result = endpoint.getOutputCluster(clusterId);
+        }
+        return result;
+    }
+
+    private void printCommands(PrintStream out, ZclCluster cluster, Set<Integer> commandIds) {
+        out.println("CommandId  Command");
+        for (Integer commandId : commandIds) {
+            out.print(" ");
```
cdjackson: Why are there 3 print statement (2 prints, followed by the println)?

hsudbrock:
Thanks for the comments, Chris! The code is more or less derived from the `attsupported` command - I will check in other commands how the in/out cluster issue is handled, as well as for the issue you mention with the exceptions, to see how I can adapt or even improve here.

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleCommandsSupportedCommand.java
hsudbrock: Copied from `attsupported` command :) But you're right, this can be improved here, thanks.

cdjackson:
See [this command](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/master/com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleReportingConfigCommand.java).

I agree - it's not standard, but I've recently come across this, and I think it's probably a good time to try and come up with a common way to handle this issue before we have too many more commands.

Maybe there are other ways to do this - eg in the same argument, we could use a colon to delimit (eg IN:0x07, OUT:1234). Something like this would make it easier to have a single method that doesn't rely on more than one argument.

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleCommandsSupportedCommand.java
hsudbrock: I agree that it is nicer to throw an exception than to simply print out a message on the console.

hsudbrock:
Chris, I have pushed a change, with an added `getCluster` method in the `ZigBeeAbstractConsoleCommand`, which takes an endpoint and a `clusterSpecifier` as parameter.

The cluster specified can be a simple integer (decimal or hex), and can be optionally prefixed with 'in:' or 'server:' (for specifying input cluster), or with 'out:' or 'client:' (for specifying an output cluster). I think it is nice to support both in/out and server/client, as I have seen both variants being used at different places.

What do you think? If this also meets what you thought of, I could adapt the code to use this from other commands as well (and make the prefixes case-insensitive, as this will otherwise be annoying quite often I believe).

cdjackson:
Thanks Henning - this looks exactly like what I was thinking, so I'm more than happy with this if you want to finish it as you suggested.

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleAbstractCommand.java
```diff
@@ -90,17 +75,77 @@ protected ZigBeeEndpoint getEndpoint(final ZigBeeNetworkManager networkManager,
...
+     * <li>in:0xB
+     * <li>server:11
+     * </ul>
+     *
+     * @param endpoint the ZigBee endpoint to get the cluster from (must be non-null)
+     * @param clusterSpecified a cluster specified as described above (must be non-null)
+     * @return the specified cluster provided by the endpoint or null if no such cluster is found
+     * @throws IllegalArgumentException if the clusterSpecifier uses an invalid number format, or if no cluster is found
+     */
+    protected ZclCluster getCluster(final ZigBeeEndpoint endpoint, final String clusterSpecifier)
```
cdjackson: clusterSpecifier doesn't match with the javadoc (clusterSpecified)

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleAbstractCommand.java
```diff
@@ -217,6 +262,16 @@ protected String printZclDataType(ZclDataType dataType) {
         return String.format("%-25s", dataType);
     }
 
+    /**
+     * @param cluster a ZCL cluster
+     * @return a String containing information about the cluster, example 'OnOff server cluster 0x00A4'
+     */
+    protected String printCluster(ZclCluster cluster) {
+        String typePrefix = cluster.isServer() ? "server " : "client ";
```
cdjackson: Are the spaces needed here? It will add an extra space in the output and therefore won't be what you list above.

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleAbstractCommand.java
```diff
@@ -90,17 +75,77 @@ protected ZigBeeEndpoint getEndpoint(final ZigBeeNetworkManager networkManager,
...
+            throws IllegalArgumentException {
+        boolean isInput;
+        boolean isOutput;
+        String clusterIdString;
+
+        if (clusterSpecifier.contains(":")) {
+            String prefix = clusterSpecifier.substring(0, clusterSpecifier.indexOf(':'));
+            isInput = prefix.equalsIgnoreCase("in") || prefix.equalsIgnoreCase("server");
+            isOutput = prefix.equalsIgnoreCase("out") || prefix.equalsIgnoreCase("client");
+            clusterIdString = clusterSpecifier.substring(clusterSpecifier.indexOf(':') + 1);
```
cdjackson: There is nothing here to check that ```in|out|client|server``` are specified - it's perfectly valid to specific ```outx:123``` and there will be no error, but you will get the input cluster (if there is one) when you are expecting the output cluster. I think if the specifier is not as expected, we should throw an invalid arg exception?

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleAbstractCommand.java
```diff
@@ -217,6 +262,16 @@ protected String printZclDataType(ZclDataType dataType) {
         return String.format("%-25s", dataType);
     }
 
+    /**
+     * @param cluster a ZCL cluster
+     * @return a String containing information about the cluster, example 'OnOff server cluster 0x00A4'
```
cdjackson: Can I suggest changing the format slightly... Instead of ```OnOff server cluster 0x00A4``` -:

```
Reading OnOff server cluster 0x00A4
Supported attributes for OnOff server cluster 0x00A4
Supported generated commands for OnOff server cluster 0x00A4
Attribute XXX was not found in OnOff server cluster 0x00A4
```

How about ```server cluster OnOff (0x00A4)``` -:
```
Reading server cluster OnOff (0x00A4)
Supported attributes for server cluster OnOff (0x00A4)
Supported generated commands for server cluster OnOff (0x00A4)
Attribute XXX was not found in server cluster OnOff (0x00A4)
```

I just think it reads better?

com.zsmartsystems.zigbee.console.main/src/main/java/com/zsmartsystems/zigbee/console/main/ZigBeeConsole.java
```diff
@@ -293,7 +292,7 @@ public void start() {
      * Processes text input line.
      *
      * @param inputLine the input line
-     * @param out the output stream
+     * @param out       the output stream
```
cdjackson: I guess you have a different formatter that is adding some spaces after parameters - or at least not removing them?

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleAbstractCommand.java
hsudbrock: oops, will fix...

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleAbstractCommand.java
hsudbrock: I agree, this is a good idea. I will make the method throw an invalid arg exception in this case.

com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleAbstractCommand.java
hsudbrock: I agree, this reads more easily.

com.zsmartsystems.zigbee.console.main/src/main/java/com/zsmartsystems/zigbee/console/main/ZigBeeConsole.java
hsudbrock: Hm, don't know how this happened; I just ran the formatter, and it removed the spaces. Thanks for the hint.

hsudbrock:
Thanks Chris, I updated the PR.

cdjackson:
Thanks Henning. Looks good.

PR #402, opened 2018-09-21
XBee does not return the channel, but, a channel mask. 

otaviojr:
XBee SC command do not returns the channel, but, a channel mask. We need to transform the mask back to the channel number in order to OH2 shows it correctly at XBee Coordinator thing.


otaviojr:
Integer.SIZE was already 32. Don't need to multiply. I will submit a new pull request.

cdjackson:
I would suggest to use the existing classes if possible (eg ZigBeeChannel, and ZigBeeChannelMask). These should allow the conversion I think - the only thing I don't remember off the top of my head is if the XBee channel mask starts with bit0=channel1, or if it is channel 11. I would prefer to reuse such classes if possible to improve readability, and reduce the chance of error.

Issue #326, opened 2018-06-16
CC2531 doesn't pass ZDO MatchDescriptor requests


cdjackson:
This requires custom code in the driver.

PR #410, opened 2018-09-29
CC2531 AF_REGISTER IasZone Cluster to Match Descriptor

otaviojr:
Chris,

I've been trying to use my Nyce Motion Sensor with the CC2531EMK dongle. Z-Stack handles the Match Descriptor Request internally, and, in order to give a successfully response we need to inform it at the AF_REGISTER command that we have an IAS Zone Cluster. Otherwise it will not be able to match the descriptor and will fail.

If you need something, please, fell free to ask.

PS.: I'm working on what you requested at the other commit. But, xstick can not handle the size of my house, and, I'm trying to make it works with CC2531EMK.

regards.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.cc2531/src/main/java/com/zsmartsystems/zigbee/dongle/cc2531/ZigBeeDongleTiCc2531.java
+++ b/com.zsmartsystems.zigbee.dongle.cc2531/src/main/java/com/zsmartsystems/zigbee/dongle/cc2531/ZigBeeDongleTiCc2531.java
@@ -420,7 +420,7 @@ public class ZigBeeDongleTiCc2531
 
         AF_REGISTER_SRSP result;
         result = networkManager.sendAFRegister(
-                new AF_REGISTER(endpointId, profileId, (short) 0, (byte) 0, new int[] {}, new int[] {}));
+                new AF_REGISTER(endpointId, profileId, (short) 0, (byte) 0, new int[] {}, new int[] {0x500}));
         // FIX We should retry only when Status != 0xb8 ( Z_APS_DUPLICATE_ENTRY )
         if (result.getStatus() != 0) {
             // TODO We should provide a workaround for the maximum number of registered EndPoint


cdjackson:
Hi,
I'm a little surprised that the ZNP stack doesn't pass these ZDO messages through. We have users who are using this with SmartThings devices which also have the same requirement. Are you absolutely certain that the ZDO messages are not passed to the application? If it didn't work, then no-one would be able to use the SmartThings devices either, and I'm sure I would have a lot of complaints.

I prefer to handle this in the application if possible as it is then uniform for all dongle types. Only if it's really not passing these messages should we do this in the dongle driver.

If the ZNP stack is really not passing this data, then I also prefer not to hard-code the clusters like this. There is a config option to allow the application to set the supported clusters, and this should be used. See the Telegesis dongle for an example.

Can you also add the ```signed-off-by``` tag to your commits (eg ```Signed-off-by: Chris Jackson <chris@cd-jackson.com>```) and ideally to the PRs as well.

Thanks.

otaviojr:
Chris, ZDO messages are going to application. The Match Descriptor Request is the only one that is not. Probably, people using CC2531, will not be able to add IAS Zones devices. Motion sensors, for example.

Match Descriptor Request is send by those devices to the coordinator and it is not a response. Z-Stack passthrough responses to the host. This request is handled by Z-Stack and if we not inform Z-Stack of our supported cluster it will fail.

I've tested a lot. And this is it. I will look forward to improve this code.

regards

cdjackson:
Ok, thanks. I will also run a test here over the next day or so as I’m just a bit surprised I’ve not seen this before.

Can you confirm what stack version you are using - just in case it’s related to that.

Thanks.

otaviojr:
If you get a look at Z-Stack ZNP documentation. You will see that the Match Descriptor Request is a message to generate the command and not to receive the command.

Every message that can be forward to the host have a note to register for receiving it. All Responses, Device Announcement, etc..

Match Descriptor Request do not have this not. Probably because will not be forwarded to host.

regards

cdjackson:
I just reminded myself that I raised this as an issue a while back -:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/326

If you can clean this up and use the standard API calls to set the clusters, that would be great.

otaviojr:
Hi Chris,

I came up with those changes. What do you think?

Am I on the right track?

regards

otaviojr:
Chris, I did it. Could you check please?

I also made PR at the bindind to use this and register the cluster.

regards

cdjackson:
Thanks - I will take a look.

Note that the binding change won't work until the libraries are updated (obviously ;) ), but this may not be for a month or two so you will need to manage the library manually.

com.zsmartsystems.zigbee.dongle.cc2531/src/main/java/com/zsmartsystems/zigbee/dongle/cc2531/ZigBeeDongleTiCc2531.java
```diff
@@ -419,8 +450,8 @@ private int createEndPoint(int endpointId, int profileId) {
         logger.trace("Registering a new endpoint {} for profile {}", endpointId, profileId);
 
         AF_REGISTER_SRSP result;
-        result = networkManager.sendAFRegister(
-                new AF_REGISTER(endpointId, profileId, (short) 0, (byte) 0, new int[] {}, new int[] {}));
+        result = networkManager.sendAFRegister(new AF_REGISTER(endpointId, profileId, (short) 0, (byte) 0,
+                this.supportedInputClusters, this.supportedOutputClusters));
```
cdjackson: I don't think we need to have ```this``` here as there are no local vars with the same name.

com.zsmartsystems.zigbee.dongle.cc2531/src/main/java/com/zsmartsystems/zigbee/dongle/cc2531/ZigBeeDongleTiCc2531.java
```diff
@@ -198,6 +213,22 @@ public void updateTransportConfig(TransportConfig configuration) {
         }
     }
 
+    private ZigBeeStatus setSupportedInputClusters(ArrayList<Integer> supportedClusters) {
+        this.supportedInputClusters = new int[supportedClusters.size()];
```
cdjackson: You can probably simplify this with something like -:

```
       supportedInputClusters = supportedClusters.stream().mapToInt(Integer::intValue).toArray();
```

com.zsmartsystems.zigbee.dongle.cc2531/src/main/java/com/zsmartsystems/zigbee/dongle/cc2531/ZigBeeDongleTiCc2531.java
```diff
@@ -198,6 +213,22 @@ public void updateTransportConfig(TransportConfig configuration) {
...
+        for (int i = 0; i < supportedClusters.size(); i++) {
+            this.supportedInputClusters[i] = supportedClusters.get(i);
+        }
+        return ZigBeeStatus.SUCCESS;
+    }
+
+    private ZigBeeStatus setSupportedOutputClusters(ArrayList<Integer> supportedClusters) {
+        this.supportedOutputClusters = new int[supportedClusters.size()];
```
cdjackson: Same comment as above...

PR #440, opened 2018-11-25
Perform critical Telegesis initialisation prior to performing any transactions

### LINKED ISSUE #435: Telegesis can not be initialized
The Telegesis dongle can't be intilialized anymore. In the logs there is one noticeable entry which says: 
`Error setting Telegesis port echo`

Probably this issue is the result of the PR #360 

[telegesis-logs.txt](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/files/2580942/telegesis-logs.txt)

cdjackson:
This adds a method to send all commands required to complete transactions. This is then called before the stick is configured, and also again during the initialisation in case the stick was reset.

This also holds off the periodic status polling if a command is currently outstanding.

I tested this by setting all 3 of these configurations incorrectly, and the dongle correctly recovers.

@TomTravaglino I hope that this will resolve your issue - please can you test this ASAP as I'd like to flow this into the latest OH binding which will be released quite soon.  Thanks.

Fixes #435 

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/ZigBeeDongleTelegesis.java
+++ b/com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/ZigBeeDongleTelegesis.java
@@ -343,6 +343,8 @@ public class ZigBeeDongleTelegesis
         frameHandler.start(serialPort);
         frameHandler.addEventListener(this);
 
+        initialiseCriticalConfiguration();
+
         // Software reset the stick so we are in a known state on startup!
         // For the first command, we add a retry mechanism
         // in case the serial port takes some time to get started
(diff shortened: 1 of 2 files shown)


TomTravaglino:
I tested the change and unfortunately it didn't help.

[telegesis-logs.txt](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/files/2619123/telegesis-logs.txt)

hsudbrock:
For the record: I tested the change as well (using openHAB), with a dongle where I manually disabled the prompt via minicom before - and for this dongle, the fix in this PR helped.

cdjackson:
The log from @TomTravaglino looks like it is working - certainly it is no longer hanging on startup, and the complete initialisation sequence seems to run so I think the original issue is resolved in this log.

@TomTravaglino why do you say it doesn't help? What is actually happening now?

hsudbrock:
I had a look at @TomTravaglino's logfile, and I think that it shows a different issue than the one that Chris tries to fix here.

The log shows that there is error 91 for the `AT+CCHANGE` command for setting the channel. Error 91 indicates `Operation only possible if connected to a PAN`.

I think that the problem here is that `ZigBeeDongleTelegesis#initialiseNetwork` tries to set the ZigBee channel (line 601) before creating the PAN (line 603 ff.).

I think that the correct way would be to not set the channel before creating the PAN, but rather to only set the channel mask to the desired channel before creating the PAN.

Does this sound right to you, Chris?

cdjackson:
I would need to check to see if this is allowed, but I would be surprised if it stopped the dongle starting. If I set the channel to a value that is not allowed (or during a time that is not allowed), and this is rejected, then it should remain at the default, and should still work ok.

I've not looked at the log in huge detail, but I'm inclined to merge this issue here as I think it solves the problem that was raised in #435 where it was not possible to communicate at all with the dongle, and we track the channel error separately (I will move the setting of the channel to another place).

WDYT?

hsudbrock:
I agree, Chris, that this PR here could be merged, and that the other issue of @TomTravaglino should be tracked in another issue.

hsudbrock:
> I would need to check to see if this is allowed, but I would be surprised if it stopped the dongle starting.

I also think that it does not stop the dongle starting; in version 1.1.2  (the version currently used by openHAB) it used to set the network status to `OFFLINE` - maybe that is the version with which @TomTravaglino tested (can you confirm this, @TomTravaglino ?). In the master branch, this has changed (with https://github.com/zsmartsystems/com.zsmartsystems.zigbee/commit/4177232e4c2a2d0ecb642474e39e4d1d2e19686e) - so the issue might be resolved by now.

TomTravaglino:
I tested with 1.1.2 and just replaced `com.zsmartsystems.zigbee.dongle.telegesis` assuming that it would be enough for a test.

hsudbrock:
Thanks, @TomTravaglino - that explains the second issue you saw (the current `master` branch would not have shown exactly the issue that I saw in your logs).

PR #453, opened 2018-12-04
Fix Transaction synchronizes on wrong variable

triller-telekom:

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransaction.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransaction.java
@@ -62,7 +62,7 @@ public class ZigBeeTransaction implements ZigBeeCommandListener {
             final ZigBeeTransactionMatcher responseMatcher) {
         this.command = command;
         this.responseMatcher = responseMatcher;
-        synchronized (command) {
+        synchronized (this.command) {
             transactionFuture = new ZigBeeTransactionFuture();
 
             // Schedule a task to timeout the transaction


triller-telekom:
While trying to find the reason for

```
Exception in thread "pool-32-thread-13" java.lang.NullPointerException
        at com.zsmartsystems.zigbee.transaction.ZigBeeTransaction.commandReceived(ZigBeeTransaction.java:109)
        at com.zsmartsystems.zigbee.internal.ZigBeeCommandNotifier$1.run(ZigBeeCommandNotifier.java:58)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

I noticed that there was a synchronization on the wrong variable, i.e. the send method synchronized on the local parameter and not on the class variable like the receive method.

I think this is a bug, but not sure if it also fixes the above mentioned NPE, as I am not 100% aware of the lifecycle of such a transaction.

cdjackson:
Thanks Stefan. Shouldn't this be the same?

```
this.command = command;
synchronize(command) {
}
```

and

```
this.command = command;
synchronize(this.command) {
}
```

Both should synchronise on the same object shouldn't they?

triller-telekom:
Alright, but then we still have to fix the NPE in

`com.zsmartsystems.zigbee.transaction.ZigBeeTransaction.commandReceived(ZigBeeTransaction.java:109)`

Lets see if it still occurs and if it does, we can fix it. Thanks for merging.

cdjackson:
Agreed. I've not looked at the exception at all yet so was just commenting on the proposed change.  Is the exception something you see often? I don't think I've ever seen this, so I'm wondering what is different if you are seeing this often?

triller-telekom:
I have seen it yesterday a couple of times and also today at least 3 times.

If it occurs again, I create an issue where I describe when it happened, i.e. what I have done before it occurred. Because right now I don't remember what I have done :)

PR #455, opened 2018-12-06
EmberStick using ASH protocol goes offline/online if stick is re-plugged

### LINKED ISSUE #319: Provide timeout handling for Ember dongle
The current implementation for the Ember dongle doesn't recognize when the controller has stopped communicating. The implementation of such a feature is under progress for the Telegesis dongle. Please, provided it also for the Ember dongle.

triller-telekom:
Implementation is based on the one for the Telegesis, see

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/322

Fixes #319


cdjackson:
Thanks Stefan,
I will take a better look over the next couple of days, but have a couple of comments from my first quick look-:

1) Can you please sign the CLA?
2) The problem I expect from this, and it is not resolved in the Telegesis implementation which you used as the reference, is that there will now be excessive logging. This is something that Henning and myself discussed for the SPI driver, and I think we will likely need to do the same sort of conditional logging to avoid the polling causing the logs to get "polluted". While this isn't solved in the Telegesis dongle, I would prefer it to be implemented for the Ember ASH as it was already for the Ember SPI.

triller-telekom:
Hi Chris,

1. In theory I can, In practice the CLA states that if I do contributions during my work time, my employer has to sign a CLA as well. So I am currently discussing with my boss whats the way forward. Maybe I have to use my personal Github account...
2. By polluting the logs you mean that since we now schedule a packet once every second, this will be printed out in the logs together with the regular traffic, right? One simple approach to this would be to simply NOT log these special `NetworkStateRequests` at all. And IF we need to see them in a rare use-case, we can still log it by ourselves. Would that be a quick and working solution?

cdjackson:
> One simple approach to this would be to simply NOT log these special NetworkStateRequests at all.

Yes, that's fundamentally what we did, but it's not so simple, right? You need to somehow know that the packet is this poll, and then conditionally log it.

triller-telekom:
@cdjackson I have made an attempt to omit the logging of status requests to debug level.

It works on my machine, BUT I still have a `magicNumber` in there which I could not map to one of your constants. Maybe you can have a look at the last commit and point me to where this `126` in a data buffer is coming from :) Thanks!

cdjackson:
Thanks. I've not looked at this properly yet, but one question... I note that the polling is done in the Ash layer - not at the dongle layer. This will therefore not work with the SPI system - I know that there is other polling for other reasons in SPI, but I still wonder if it might not be better done at a higher level (ie in the dongle).

com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java
```diff
@@ -461,10 +505,35 @@ private void sendRetry() {
...
     }
 
+    private boolean isStatusRequest(AshFrame frame) {
+        int magicNumber = 126;
+        int[] buffer = frame.getOutputBuffer();
+
+        int[] rawDataBuffer = frame.dataBuffer;
+
+        // AshFrameAck
+        if (buffer.length == 4 && buffer[3] == magicNumber) {
```
cdjackson: Are you sure this works across all versions of ASH? I need to look at this more, but why are you peeking into the raw buffer rather than the frame type? Maybe it's not decoded at that point? I think it would be much better to use the decoded frame, otherwise I don't think this code works with the different versions of ASH.

triller-telekom: Yes, right. Its not decoded at this point. If I do `frame.getOutputBuffer()` I get nothing for this type (AshFrameAck) of paket.

That's why I was asking for your help, since its your code :)

Regarding the layer issue: I thought that chips connected via SPI cannot be detached anyway, so there is no need for such a heartbeat. Also I looked at the types of FrameHandlers and there is the one for SPI and the one for ASH, they do not have a common Base class. Can you point me to the EZSP "frame handling" class where you imagine the  heartbeat should be placed?

But anyhow, if you cannot detach a SPI chip, do you think we need that heartbeat? Because all we could tell the user is: "Your chipset is broken, it doesn't respond"...

cdjackson:
> Yes, right. Its not decoded at this point.

I think this needs to use the standard systems - somehow! At the moment there are two different implementations of ASH, but there may be more in future, and at some point I will probably need to add security. I've not looked at that yet, but I suspect it might also break the checks here.

> Regarding the layer issue: I thought that chips connected via SPI cannot be detached anyway, so there is no need for such a heartbeat.

I'm not sure I would agree - if we're adding the check, in my mind it's not just to make sure people don't remove the dongle. UART dongles running ASH are also cannot be detached in many systems. For me, we want to know if the NCP dies for any reason (firmware crash, hardware failure, removal from USB...). I think if we are going to implement something, it should cover these cases.

> But anyhow, if you cannot detach a SPI chip, do you think we need that heartbeat? Because all we could tell the user is: "Your chipset is broken, it doesn't respond"...

I think we need to know this. The fact that it might be a hardware failure is not relevant - they want to know if the system is working and they can rely on being able to use it don't they? That's the important point I think?

triller-telekom:
>> Yes, right. Its not decoded at this point.

> I think this needs to use the standard systems - somehow! At the moment there are two different implementations of ASH, but there may be more in future, and at some point I will probably need to add security. I've not looked at that yet, but I suspect it might also break the checks here.

Since at that point I only have the `AshFrame`, can you please point me to some code where I can decode it with? Or does this decoding code simply not exist yet?

In the `rawBuffer` I can check for `rawDataBuffer[2] == EzspNetworkStateRequest.FRAME_ID`, so I know it is the NetworkStateRequest frame. However if I receive an `AshFrameAck` frame, I don't know how to tell the frame type from it.

>> But anyhow, if you cannot detach a SPI chip, do you think we need that heartbeat? Because all we could tell the user is: "Your chipset is broken, it doesn't respond"...

> I think we need to know this. The fact that it might be a hardware failure is not relevant - they want to know if the system is working and they can rely on being able to use it don't they? That's the important point I think?

Alright, you have convinced me that this might be useful to **some** users. I have added a commit which moves the logic one layer up into the EZSP layer.

So the open question which remains is the one for filtering out the log messages. IMHO this isn't very critical to me, so any solution which removes them for the logs is fine for me. If you could point me to some code for decoding so I can improve it, I would very much appreciate it, see above.

cdjackson:
> Since at that point I only have the AshFrame, can you please point me to some code where I can decode it with? Or does this decoding code simply not exist yet?

The code exists, but at the point in the code you are trying to use it, maybe it has not been decoded. The problem is that there are 2 ASH versions (currently). You need to know what one is being used or you will not decode the frame correctly. I don't think your code, as currently written, will work on more recent versions of EZSP as the byte that you are looking at is in a different location. In future, there are likely to be more changes as there is also a secure implementation. If you are going to peek into the frames, then it gets messy here as you will need to handle all these options.

> In the rawBuffer I can check for rawDataBuffer[2] == EzspNetworkStateRequest.FRAME_ID, so I know it is the NetworkStateRequest frame. 

Again though, I think this check will not work every in all instances - what version of EZSP are you verifying against?

To illustrate my point, below is the frame definition for different versions of EZSP -:
 
image

image

You are looking at the frameId byte and as you can see, it is not in the same place in these two different definitions.

> If you could point me to some code for decoding so I can improve it, I would very much appreciate it, see above.

I'm not sure what you mean here? The code for decoding the frames is in the AshFrame implementations. The issue might be that we need to look at where data is logged so that we only log at debug or above AFTER the frame is decoded.

triller-telekom:
I understand that there are multiple versions of ASH (just learned that though) and I admit that my picking of byte positions is kind of hacky...

> at the point in the code you are trying to use it, maybe it has not been decoded

Exactly that's my problem. But AFAIK int he code there is no way to get the FrameID from a `AshFrameData` or `AshFrame` object which I have at the time it is logged. Or am I blind?

An `AshFrame? simply only has a `getFrameType()` method, bit this only gives you:

```
enum FrameType {
        DATA,
        ACK,
        NAK,
        RST,
        RSTACK,
        ERROR
    }
```

What I meant with "please point me to the code" is, that you say there IS code to decode the frame, so I am looking for that code which decodes it into a form where I have an object which I can ask something like `getFrameID()`.

cdjackson:
> Exactly that's my problem. But AFAIK int he code there is no way to get the FrameID from a AshFrameData or AshFrame object which I have at the time it is logged. Or am I blind?

Probably you are right (I've not checked the code). This is my point that we may have to look at reducing some logging to trace where the logging occurs before the frame is decoded. I will have a look/think over the weekend.

> you say there IS code to decode the frame

Yes - clearly that is the case or the frame would never be decoded :)

What you are looking for is in the Ezsp frame decoding - not the ASH frame decoding.

> I am looking for that code which decodes it into a form where I have an object which I can ask something like getFrameID().

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/000375438763f4b1d5a168548aa388e43669fd20/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ezsp/EzspFrame.java#L379

triller-telekom:
> This is my point that we may have to look at reducing some logging to trace where the logging occurs before the frame is decoded. I will have a look/think over the weekend.

Did you think about it meanwhile?

Regarding the CLA: For this PR I will squash the commits, change the sign-off to my private email and sign the CLA, once we have an agreement on the logging.

cdjackson:
Sorry - I’ve not looked at this too much as I spent most of the weekend preparing stuff for the OH 2.4 release.

One thing I did “conclude” though - I don’t think we need to reduce logging in the ASH layer to TRACE. I think this can stay as it is, and if people don’t want to have ASH level debug logging, then they should just not log that layer. So I think we should leave ASH layer alone, and focus on the EZSP layer where we don’t want to log these messages as often as they are being sent.

I will try and look at this later in the week.

triller-telekom:
> I don’t think we need to reduce logging in the ASH layer to TRACE. I think this can stay as it is, and if people don’t want to have ASH level debug logging, then they should just not log that layer.

Thinking about this again... What if one wants to see the debug logs of the ASH layer, but without the "noise" of the periodically send EZSP status requests? I.e. the EZSP status requests introduced with this PR result in ASH packets which then get logged.

Or do you say that if one wants to see ASH packets one usually also wants to see these status requests?

cdjackson:
> Or do you say that if one wants to see ASH packets one usually also wants to see these status requests?

Exactly. If someone is debugging the ASH layer, then they want to see ALL ASH data - not just some of the packets since that won't allow the protocol to be debugged as there would be many missing packets in the sequence.  If you don't want to see the ASH protocol, then don't enable ASH debug.

If we lower this to TRACE, then I think someone wanting to DEBUG ASH, will always need to run this layer in TRACE, which defeats the purpose IMHO.

PR #468, opened 2018-12-15
Fixed ASH retry counter and added tests

### LINKED ISSUE #466: ASHFrameHandler does not handle retries correctly
As a follow-up to:
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/464#discussion_r241715960

the `ASHFrameHandler` starts a Thread for each request, but all seem to share the same `retries` field, which gets reseted whenever a new Thread is created.

Thus, the maximum retry limit is never reached and the `disconnect` won't be called to announce that the stack is OFFLINE.

cdjackson:
@triller-telekom FYI this ought to resolve your problem.
Closes #466 


cdjackson:
https://github.com/openhab/org.openhab.binding.zigbee/pull/335

PR #485, opened 2018-12-24
Typo for gradle and version changed

milansamardzic:


README.md
```diff
@@ -283,7 +283,7 @@ The repositories below may be used for the RELEASE and SNAPSHOT versions respect
 <dependency>
     <groupId>com.zsmartsystems.zigbee</groupId>
     <artifactId>com.zsmartsystems.zigbee</artifactId>
-    <version>1.1.5</version>
+    <version>1.x.x</version>
```
cdjackson: I think this should still be 1.x.x - it was simply meant to denote to use the latest version, or whatever version the user likes. 1.1.5 is not the most recent if that was your intention, and in any case, we should make this consistent with the dependency which you left as 1.x.x.

My preference is to leave this as 1.x.x - I think that most people should understand that this needs to change. If you think that's not obvious, then I would suggest to leave as 1.x.x, but make a comment that this needs to be set to the version the user wants to use.

milansamardzic: @cdjackson 
I changed from 1.1.5 to 1.x.x in this commit, because I found that 1.1.7 is currently latest version. 
I will add comment in readme with a suggestion to change x.x with the latest/desired version.

cdjackson:
Thanks

PR #495, opened 2019-01-04
Don't poll ASH state when ASH layer is OFFLINE

### LINKED ISSUE #494: AshFrameHandler should stop queuing data if ASH layer goes from ONLINE to OFFLINE
This is a follow-up on the discussion done here:
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/466#issuecomment-451427278

If the ASH layer goes down, the EZSP layer will still continuously send out its status request packets. Those packets are queued up and fill up the device memory.

To solve this, the idea in the linked discussion is to reject incoming TX requests from upper layers towards the ASH layer if the ASH was in ONLINE state and is in OFFLINE state now. The part "was in ONLINE state" is important, because during initialization of the whole zigbee stack we do want the queuing although the ASH layer is not yet ONLINE.

cdjackson:
This makes two changes relating to the polling of the ASH layer.

1) It doesn't poll the state if the ESZP layer is OFFLINE. This avoids any polling during startup, and also when the ASH layer has closed.
2) If there is an error in the ASH layer, the ASH layer is closed and no further transactions are accepted. Once there is an error, the upper layers need to reinitialise the NCP, so this should have no negative impact.

Closes #494 


cdjackson:
@triller-telekom if you have any comments please let me know and we can discuss, but I think this ought to resolve any queueing in the ASH layer as it is now closed following any error :)

PR #498, opened 2019-01-08
WIP: Add support for manufacturer-specific clusters

hsudbrock:
Hi @cdjackson; we discussed last week how to make it possible to extend the framework with manufacturer-specific clusters, in particular about the option to introduce a registry for clusters that provides (a) manufacturer-specific clusters that have been added to the registry by an application that integrates this library, and (b) the standard clusters from the cluster library that are currently provided by the enum `ZclClusterType`.

For starters, I have implemented this approach for `profile types`, because those are somewhat simpler than the clusters, while the general approach is sufficiently similar:

* The profile type enum with the profile types provided by the framework is renamed to `StandardZigBeeProfileTypes`
* This enum implements the new interface `ZigBeeProfileType`.
* The network manager provides a `ZigBeeProfileTypeRegistry`, to which manufacturer-specific profile types can be added; lookup of profile types is no longer via the enum, but via the registry. The registry will return both the manufacturer-specific profile types, and the profile types from the autocoded enum.

I think that this approach can be applied likewise to the clusters (renaming the `ZclClusterType` enum to `StandardZclCllusterType`, introducing an interface for cluster types and make the enum implement the interface, and add a `ZclClusterTypeRegistry` to the network manager, via which manufacturer-specific clusters can be added by clients of the network manager).

If you already have any comments / suggestions on this first step, I'd be happy to hear.

P.S. I will add a PR that manages the manufacturer-specific bits in `ZclFrame` serialization separately.


cdjackson:
Thanks Henning. I think I would prefer to see a larger PR that showed why you are changing this enum. I guess it is since it is used in the cluster type list, but at the moment with just this PR, it's unclear.

My only other comment so far is that instead of ```StandardZigBeeProfileType```, I would prefer ```ZigBeeStandardProfileType```.

hsudbrock:
Hi Chris; no problem, I will extend the PR.

I posted this PR already now merely to illustrate the approach of using an interface instead of one of the existing enums, having the existing enum implement the interface, and having a registry to make it possible to add manufacturer-specific instances of the interface.

I asked myself the question whether it is actually worth to permit adding manufacturer-specific profiles as well (for us, manufacturer-specific clusters and attributes are what we actually need), but I thought that it is sensible to do this, as one of the manufacturer-specific clusters comes with a manufacturer-specific profile as well (though I could use the library with that particular cluster also without adding a manufacturer-specific profile, as the library probably just passes the profile ID around, and does not require the profile to be known).

hsudbrock:
@cdjackson One point where I'd be happy to get your opinion:

We discussed to make the enum for ZCL standard cluster types implement an interface for cluster types (so that this interface can be implemented by non-standard cluster types as well). The obvious name for this interface would be `ZclClusterType` - but this is currently the name of the enum. If we rename the enum (e.g., to something like `ZclStandardClusterType`), this is an API-breaking change (requiring changes in consumers of the library, like the openHAB binding).

I, personally, prefer doing the API-breaking change by renaming the enum, and adapt the openHAB binding accordingly - but I don't know how big the impact on other consumers of the library would be. If we do not want to rename the enum, we need to find a 'good enough' name for the interface - but I did not find anything really convincing (the best being `IZclClusterType` - though I am not a big fan of adding the `I` prefix for indicating an interface).

WDYT?

cdjackson:
Thanks Henning. Let's shoot with renaming the classes. I think this should be ok, but if it really looks too major when we come to merge we could always rename them. I am in general though inline with you.

hsudbrock:
I have updated the PR by adding support for manufacturer-specific cluster types:
* The enum `ZclClusterType` is renamed to `ZclStandardClusterType`
* The enum implements the new interface `ZclClusterType`
* The `ZclClusterTypeRegistry` is the entry point for retrieving cluster types; it permits to add manufacturer-specific cluster types, and will provide those as well as the cluster types from `ZclStandardClusterType`.
* The `ZigBeeNetworkManager` will now set the manufacturer code and the manufacturer-specific bit to commands that come from a command marked as 'manufacturer-specific'; in particular, the `ZclCluster` class will mark commands as manufacturer-specific if the cluster is manufacturer-specific.

I have implemented the registries as singletons; I had played around with adding the registries to the `ZigBeeNetworkManager`, so that it is possible to provide different sets of manufacturer-specific clusters for different networks, but rejected this idea for simplicity, in particular as access to the registries is needed from parts of the code from which the reference to the network manager was just removed (e.g., the `ZigBeeEndpoint` class).

As discussed, I have added the (small) changes to the autocoder for now, being aware that I might have to migrate this after ongoing changes to the autocoder.

I will add a WIP PR for the ZigBee binding in a couple of minutes which illustrates how this new infrastructure for manufacturer-specific cluster could be used.

I have only added support for adding manufacturer-specific clusters for now, and would add support for adding manufacturer-specific attributes and commands later, but think that this could be a good state to start discussion here.

cdjackson:
@hsudbrock when we discussed this the other day I was a bit surprised that you'd not come across the ```ZclCommandType``` enum as I think receiving unknown (eg manufacturer specific) commands will not work until this enum is changed. Or are you currently manually adding all your commands to this enum - which of course would also not be a long term solution.

For discussion... my thinking at the moment is to change the way that commands get instantiated when the command is received. Currently, this is all done in the manager as there is this "nice" enum that lists "all" the commands. However, that is not the full story and the current implementation will not work with manufacturer specific commands.

I see two options.

1) Do the same thing for commands that you have done for clusters and have a list that can be added to by the application. I don't really like this, and I don't think it will work - or at least there may be the case that different manufacturers add the same mfgr command to the same cluster, so we'd need to handle this.
2) Instead we move the instantiating of commands into the clusters themselves. In fact this is partly already there with the ```getCommandFromId``` and ```getResponseFromId``` methods. Clearly the current implementation is not the right approach as it's not extendable, and possibly a ```Map``` of commands would be better.  This is my current preferred approach as it makes clusters responsible for their commands rather  than using a global method which doesn't really feel right.

WDYT? I was looking at implementing the above in the new XML coder so that it's available for this PR, but wanted to get your thoughts first...

hsudbrock:
> when we discussed this the other day I was a bit surprised that you'd not come across the `ZclCommandType` enum as I think receiving unknown (eg manufacturer specific) commands will not work until this enum is changed. Or are you currently manually adding all your commands to this enum - which of course would also not be a long term solution.

I think the reason I did not come across this is that I have so far only dealt with sending manufacturer-specific commands (simply because for the devices with manufacturer-specific clusters that I have here I had to sent commands out, and not vice versa).  So you are completely right, this is not resolved so far and needs to be resolved, thanks for pointing this out.

> WDYT? I was looking at implementing the above in the new XML coder so that it's available for this PR, but wanted to get your thoughts first...

I would also prefer option 2, i.e., making the clusters responsible for their commands. This would then also be somewhat in line with attributes, which are currently also managed by the cluster classes themselves. I think that it would be good to have similar structures for attributes and for commands, so that the extension mechanism for those two could also be the same, so using a map as you suggest sounds good.

cdjackson:
This pull request **introduces 3 alerts** and **fixes 8** when merging 759213f1b2be8ac96b42d6b36731455dae932889 into ec8c99e3a715b374b1cf332f1865bb7ad6ed87d5 - [view on LGTM.com](https://lgtm.com/projects/g/zsmartsystems/com.zsmartsystems.zigbee/rev/pr-7fa714b3a5f63d3aac0da920a7cb016c9c0d3143)

**new alerts:**

* 1 for Useless comparison test
* 1 for Spurious Javadoc @param tags
* 1 for Dereferenced variable may be null

**fixed alerts:**

* 7 for Useless comparison test
* 1 for Useless null check

---

cdjackson:
This pull request **introduces 3 alerts** and **fixes 8** when merging 3bade215ac73493cce2a153df6f4412226fc1224 into ec8c99e3a715b374b1cf332f1865bb7ad6ed87d5 - [view on LGTM.com](https://lgtm.com/projects/g/zsmartsystems/com.zsmartsystems.zigbee/rev/pr-c39b9f73d996d22e6cd2b26e5e397f024fe5726d)

**new alerts:**

* 1 for Useless comparison test
* 1 for Spurious Javadoc @param tags
* 1 for Dereferenced variable may be null

**fixed alerts:**

* 7 for Useless comparison test
* 1 for Useless null check

---

hsudbrock:
I have rebased this PR on the PR for the new XML Autocoder (removing the option to provide custom profile types on the way, as (a) this is not so easily possible anymore with how the autocoder generates the enum with the profile types, and (b) as it now seemed kind of out of scope in a PR for manufacturer-specific clusters to me).

The current implementation permits to provide complete manufacturer-specific clusters. Next steps are adding support for extending standard clusters with manufacturer-specific commands and attributes, but I think that this should wait until you have moved responsibility for commands from the `ZclCommandType` enum to the cluster implementations, as you described above, @cdjackson .

cdjackson:
Thanks Henning I will take a look at the refactoring for the command generation over the coming days.

cdjackson:
This pull request **introduces 1 alert** when merging c84d4e1c2891fc3f5e447cb4b2df987c52fbaa23 into a47ff1ba6d02785128d203380fac79c71e7b5921 - [view on LGTM.com](https://lgtm.com/projects/g/zsmartsystems/com.zsmartsystems.zigbee/rev/pr-e6e17f773c092611d2418a22835a58c1c6e0d95e)

**new alerts:**

* 1 for Spurious Javadoc @param tags

---

cdjackson:
This pull request **introduces 1 alert** when merging 30feb43c8a3fb3bb5d1842b00a55dd1aaadde1f7 into a47ff1ba6d02785128d203380fac79c71e7b5921 - [view on LGTM.com](https://lgtm.com/projects/g/zsmartsystems/com.zsmartsystems.zigbee/rev/pr-61a9a257fb43c1ffce367371962f0f22ddd340cc)

**new alerts:**

* 1 for Spurious Javadoc @param tags

---

hsudbrock:
As a summary, this is how I have implemented support for manufacturer-specific clusters, commands, and attributes in the current state of this PR:

1.) Extension of `ZclClusterType`, `ZclCluster`, `ZclAttribute`, and `ZclCommand` with manufacturer code

Each manufacturer-specific cluster, attribute, and command is for a specific manufacturer code, and this code is represented as an `Integer` in the corresponding classes. For non-manufacturer-specific instances, the code must be `null`.

The `ZigBeeNetworkManager` adds the manufacturer code to the `zclHeader` for manufacturer-specific commands. (When receiving a command, the `ZigBeeNetworkManager` could, in principle, check that the manufacturer code of the incoming frame matches the manufacturer code of the `ZclCommand` provided by the `ZclCluster` instance - but this is not done now as I do not see the benefit of checking this.)

For generic commands, it depends on the attributes referenced in the command whether the manufacturer code needs to be set in the `zclHeader`. This PR adds this logic in the methods of the `ZclCluster` class for sending out generic commands. For example, the method `ZclCluster.read` creates a `ReadAttributesCommand`, and it now sets the manufacturer code for this command if (a) the command is sent to a manufacturer-specific cluster, or if (b) the command is for reading manufacturer-specific attributes of a standard cluster. Unfortunately, the ZCL specification does not mention the case that a generic command references both standard and manufacturer-specific attributes (e.g., reading a standard and a manufacturer-specific attribute with one read command). This PR will set the manufacturer code in the command only if all referenced attributes are manufacturer-specific.

1.) Possibility to add manufacturer-specific clusters

The PR renames the `ZclClusterType` enum to `ZclStandardClusterType`, and introduces a new interface `ZclClusterType` which is implemented by the `ZclStandardClusterType` enum. Both standard cluster types (from `ZclStandardClusterType`) and manufacturer-specific cluster types are accessed via the singleton `ZclClusterTypeRegistry`. To add support for a specific manufacturer-specific cluster type, clients of this library can add an implementation of `ZclClusterType` via the `ZclClusterTypeRegistry`.

In `ZigBeeEndpoint`, the correct instance of `ZclClusterType` is retrieved from the `ZclClusterTypeRegistry`, using the cluster ID and the manufacturer code from the node descriptor.

2.) Possibility to add manufacturer-specific attributes and commands

Clients of this library can add manufacturer-specific attributes or commands to an instance of `ZclCluster` via the methods `addAttributes`, `addClientCommands`, and `addServerCommands`.

cdjackson:
This pull request **introduces 1 alert** when merging a2f6a567e5ac926b478968cf7d3028c0e2bbb067 into a47ff1ba6d02785128d203380fac79c71e7b5921 - [view on LGTM.com](https://lgtm.com/projects/g/zsmartsystems/com.zsmartsystems.zigbee/rev/pr-ab16847083e5bebf88ddf1d79991bbce15e6a5b2)

**new alerts:**

* 1 for Spurious Javadoc @param tags

---

hsudbrock:
@cdjackson As discussed last Friday, we want that clients of this library add manufacturer-specific commands and attributes to standard clusters by themselves (via methods called on concrete instances of `ZclCluster`). 

I am now thinking about whether we shouldn't do the same for adding manufacturer-specific clusters. When we discussed this when I started working on this a couple of weeks ago, we ended up with the idea to have some kind of "cluster type registry", to which one could add definitions of manufacturer-specific clusters (and that's also how it's currently implemented in this PR).

However, it seems more consistent to me now to use the same approach for manufacturer-specific clusters as for manufacturer-specific attributes and commands (as discussed last Friday) - i.e., letting clients of this library add them to an existing endpoint by themselves.

WDYT?

cdjackson:
Closing this as there will be a replacement PR.

PR #500, opened 2019-01-08
Add manufacturer-specific infos when serializing ZCL frames

hsudbrock:
The deserialization logic already processes the manufacturer-specific bits. This commit adds the logic for serializing ZCL frames, more specifically:

* Sets the manufacturer-specific bit in the frame control field for instances of `ZclHeader` that have the boolean value `manufacturerSpecific` set.
* Adds the manufacturer code to ZCL frames if the manufacturer-specific bit is set.


cdjackson:
Thanks Henning.

Issue #450, opened 2018-12-03
Transaction timeout while transferring the last chunk of the firmware image


hanneshofmann:
While trying to update the device firmware of the Bitron Home 902010/25, the last transaction of the `ImageBlockResponse` has a timeout and therefore, the firmware update gets aborted. Please note that we have observed that issue not only with the Bitron Home device but also with devices from other vendors (e.g. CentraLite, SengLED,..).

The following log captures the transfer of the last chunk of the firmware image:
````
2018-11-30_13:50:29:174	NA	DEBUG:	[symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - RX CMD: ImageBlockCommand [OTA Upgrade: 17784/4 -> 0/1, cluster=0019, TID=C1, fieldControl=2, manufacturerCode=10132, imageType=257, fileVersion=48, fileOffset=211140, maximumDataSize=36, requestNodeAddress=null, blockRequestDelay=0]
2018-11-30_13:50:29:175	NA	DEBUG:	[symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.app.otaserver.ZclOtaUpgradeServer] - 17784/4 OTA Data: Sending 36 bytes at offset 211140
2018-11-30_13:50:29:177	NA	DEBUG:	[symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - TX CMD: ImageBlockResponse [OTA Upgrade: 0/0 -> 17784/4, cluster=0019, TID=D9, status=SUCCESS, manufacturerCode=10132, imageType=257, fileVersion=48, fileOffset=211140, imageData=ByteArray [value=80 02 79 00 02 12 11 C0 82 C0 83 7A 00 79 00 12 23 EA D2 87 90 61 96 74 06 F0 D0 83 D0 82 02 12 11 FF FF FF]]
2018-11-30_13:50:29:178	NA	DEBUG:	[symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - TX ZCL: ZclHeader [frameType=CLUSTER_SPECIFIC_COMMAND, manufacturerSpecific=false, direction=SERVER_TO_CLIENT, disableDefaultResponse=false, manufacturerCode=0, sequenceNumber=217, commandId=5]
[...]

PR #519, opened 2019-02-01
Made OTA more robust by increasing retries and transfer timeouts

hanneshofmann:
This PR is intended to close https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/450.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
@@ -198,12 +198,22 @@ public class ZclOtaUpgradeServer implements ZigBeeApplication {
     /**
      * Default transfer timeout period in milliseconds
      */
-    private final long TRANSFER_TIMEOUT_PERIOD = 30000;
+    private static final long TRANSFER_TIMEOUT_PERIOD = 300000;
 
     /**
      * Transfer timeout period in milliseconds
      */
     private long transferTimeoutPeriod = TRANSFER_TIMEOUT_PERIOD;
+
+    /**
+     * The amount of retries to get the current firmware version
+     */
+    private static final int AMOUNT_OF_RETRIES_TO_REQUEST_CURRENT_FW_VERSION = 10;
+
+    /**
+     * The sleep time before trying to request the current firmware version
+     */
+    private static final long SLEEP_TIME_BEFORE_REQUESTING_CURRENT_FW_VERSION = 10000;
 
     /**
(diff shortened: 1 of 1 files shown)


cdjackson:
I must say that I'm very surprised that the timeout is now required to be 5 minutes and 20 retries are required to get the firmware version. This doesn't really sound right - it might solve your issue, but maybe this means there is a problem with the device? Or do you have another explanation as to why this is needed as you don't provide any information with the PR?

I'm not really happy to increase these defaults to this extent unless the protocol definition states that this is normal (it seems quite wrong to me) or there is a clear explanation. Maybe we should have a method to set these timeouts so it can be changed on a per device basis for your implementation.

hanneshofmann:
First of all, thanks for your response and sorry for not providing any further information on this PR. This PR is intended to illustrate where we currently face issues regarding OTA. 

>This doesn't really sound right - it might solve your issue, but maybe this means there is a problem with the device? 

I completely agree with you that increasing the amount of retries and the timeout might solve individual device-specifics but is wrong in general.

>I'm not really happy to increase these defaults to this extent unless the protocol definition states that this is normal (it seems quite wrong to me) or there is a clear explanation.

Honestly, I don't know whether the protcol definition states that it is normal or not. I had observed that the current timeout of 30 seconds and the amount of 10 retries were not sufficient for almost every device* (except Osram bulbs :)).

> Maybe we should have a method to set these timeouts so it can be changed on a per device basis for your implementation.

That would work. As the transfer timeout is already configurable, we do only need to introduce a setter for the sleep time or the amount of retries to request the software version after the update - what do you prefer?


*I tested with several Bitron Video, Bitron Home, CentraLite and Osram devices including main- and battery-powered.

cdjackson:
Sorry @hanneshofmann I missed the log you attached in the other issue.

I think for this issue, it is fine to increase ```TransferTimeoutPeriod``` - I think a longer default than the current 30 seconds is probably fine (the 5 minutes you propose in your PR seems reasonable). There is no timeout defined in the protocol docs (at least not that I can find). I guess that the device is taking time to verify the image has the correct checksum/hash/? before it sends the ```UpgradeEndCommand```.

The retries on the request for the version doesn't appear in the log you provided as the transfer doesn't complete. However, for the same reason as above, I think it's fair to extend the time here.

I suggest to update the retry time and max count to use a constant at the top of the file - if you want to provide setters, then feel free as well.

hanneshofmann:
Thanks for your feedback, @cdjackson.

I also suppose that the current transfer timeout is not sufficient as the most ED might require more time to check the image integrity before they send the `UpgradeEndCommand`.

As suggested, I extracted the amount of retries and the sleep time as constants. Besides that, is there something else missing in the MR? I hope you don't mind that I've squashed and force pushed the latest changes.

cdjackson:
Thanks - that looks fine to me.

Issue #525, opened 2019-02-09
BindRequest transactions do not seem to ever complete


puzzle-star:
Hi @cdjackson,

After upgrading the library and the OH binding, most of my devices do not seem to ever report attribute changes. Maybe this is happening from before, so I cannot tell exactly from when this is happening.

Looking at the log, I see that even if the BindRequest command succeeds and I get the response, this response does not match the transaction:

```
2019-02-09 09:45:38.802 [DEBUG] [ng.zigbee.handler.ZigBeeThingHandler] - 000B57FFFEA46B77: Initializing channel zigbee:device:04000AB3:000b57fffea46b77:000B57FFFEA46B77_1_dimmer with org.openhab.binding.zigbee.internal.converter.ZigBeeConverterSwitchLevel@c27612
2019-02-09 09:45:38.804 [DEBUG] [zigbee.transaction.ZigBeeTransaction] - Sending transaction: BindRequest [null -> 15773/1, cluster=0021, TID=NULL, srcAddress=000B57FFFEA46B77, srcEndpoint=1, bindCluster=8, dstAddrMode=3, dstAddress=000D6F001113A331, dstEndpoint=1] ==== BindRequest [null -> null, cluster=0021, TID=NULL, srcAddress=null, srcEndpoint=null, bindCluster=null, dstAddrMode=null, dstAddress=null, dstEndpoint=null]
2019-02-09 09:45:38.806 [DEBUG] [tsystems.zigbee.ZigBeeNetworkManager] - TX CMD: BindRequest [0/0 -> 15773/1, cluster=0021, TID=D3, srcAddress=000B57FFFEA46B77, srcEndpoint=1, bindCluster=8, dstAddrMode=3, dstAddress=000D6F001113A331, dstEndpoint=1]
2019-02-09 09:45:38.808 [DEBUG] [tsystems.zigbee.ZigBeeNetworkManager] - TX APS: ZigBeeApsFrame [sourceAddress=0/0, destinationAddress=15773/0, profile=0000, cluster=0021, addressMode=DEVICE, radius=31, apsSecurity=false, apsCounter=D3, payload=00 77 6B A4 FE FF 57 0B 00 01 08 00 03 31 A3 13 11 00 6F 0D 00 01]
[...]

PR #526, opened 2019-02-09
Fix bind response matcher issue #525

puzzle-star:
This makes the bind() and unbind() methods in ZclCluster send the commands via 

`zigbeeEndpoint.getParentNode().sendTransaction()`

instead of directly via

`zigbeeEndpoint.sendTransaction()`

To prevent the endpoind method from setting destination endpoint in the zigbee header, as the ZDO command should be addressed to endpoint 0.

This should fix issue [BindRequest transactions do not seem to ever complete #525](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/525#issue-408413902)

Tested in my OH environment and seems to be correcting the problem

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
@@ -472,8 +472,6 @@ public abstract class ZclCluster {
         command.setDstAddress(address);
         command.setDstEndpoint(endpointId);
         // The transaction is not sent to the Endpoint of this cluster, but to the ZDO endpoint 0 directly.
-        // Hence calling here zigbeeEndpoint.getParentNode().sendTransaction() instead of
-        // zigbeeEndpoint.sendTransaction()
         return zigbeeEndpoint.getParentNode().sendTransaction(command, command);
     }
 
@@ -494,8 +492,6 @@ public abstract class ZclCluster {
         command.setDstAddress(address);
         command.setDstEndpoint(endpointId);
         // The transaction is not sent to the Endpoint of this cluster, but to the ZDO endpoint 0 directly.
-        // Hence calling here zigbeeEndpoint.getParentNode().sendTransaction() instead of
-        // zigbeeEndpoint.sendTransaction()
         return zigbeeEndpoint.getParentNode().sendTransaction(command, command);
     }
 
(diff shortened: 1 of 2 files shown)


puzzle-star:
Hi @cdjackson,

I have also updated ZclClusterTest as now the bind and unbind calls are using getParentNode() on the endpoint and this was returning null on the test (and hence failing)

I had never seen anything called "Mockito" before, so if I have done something really wrong please let me know :-)

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -471,7 +471,10 @@ public boolean getApsSecurityRequired() {
         command.setDstAddrMode(3); // 64 bit addressing
         command.setDstAddress(address);
         command.setDstEndpoint(endpointId);
-        return zigbeeEndpoint.sendTransaction(command, new BindRequest());
+        // The transaction is not sent to the Endpoint of this cluster, but to the ZDO endpoint 0 directly.
+        // Hence calling here zigbeeEndpoint.getParentNode().sendTransaction() instead of
```
cdjackson: Can I ask that you remove the second and third lines of these two comments? ie I think the first line explains the point, and the code shows what is called ;)

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -490,7 +493,10 @@ public boolean getApsSecurityRequired() {
         command.setDstAddrMode(3); // 64 bit addressing
         command.setDstAddress(address);
         command.setDstEndpoint(endpointId);
-        return zigbeeEndpoint.sendTransaction(command, new UnbindRequest());
+        // The transaction is not sent to the Endpoint of this cluster, but to the ZDO endpoint 0 directly.
+        // Hence calling here zigbeeEndpoint.getParentNode().sendTransaction() instead of
```
cdjackson: Can I ask that you remove the second and third lines of these two comments? ie I think the first line explains the point, and the code shows what is called ;)

puzzle-star: Done :)

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
puzzle-star: Done

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
cdjackson: Many thanks Pedro - I'll merge once CI completes.

PR #527, opened 2019-02-09
New autocoder employing XML definition file format

cdjackson:
This PR is a rewrite of the autocoder. Primarily, it changes from an ```MD``` format to an ```XML``` format. The MD file was getting unmanageable in size as more clusters are added, and it is also more difficult to parse and add new constructs.

This may change the name of some methods due to the way the file is parsed - eg ```Iascie``` is now formatted in methods/fields as ```IasCie``` and ```ZoneID``` is now ```ZoneId``` so this may break some implementations (sorry!).

There may also be some changes to commands and attributes as these are now marked with a direction (eg wether an attribute resides in the client or server). This is one of the drivers behind this change - ie to allow a future enhancement which supports different attributes for clients and servers, as required by some clusters.

Currently this PR adds attributes for the server. This needs to be resolved as the OTA is implemented as a client on the device, so this will cause it to not work right now!

@hsudbrock FYI. I still need to test this - it generates code that is pretty much the same as before, but there might be some changes...

Signed-off-by: Chris Jackson <chris@cd-jackson.com>

PR #531, opened 2019-02-12
[WIP] execute autocode from maven

t-8ch:
As discussed in #527 . This is an early prototype to gather feedback.
The commits are structured as follows:

1) Delete all generated code, to make it easier to exclude from review
2) Should probably be applied to master right away.
3) The logic of the change.

Caveats:
* Currently only supports the base library. The dongle generation is possible as well. Maybe a dedicated mojo per project or via configurations, if they are similar enough.
* Only maven (no gradle). Is there a reason for two build systems?
* Executes the generation twice. (once triggered by the maven-source-plugin, may be disabled)
* Currently the source directory is custom and not discovered by IDEs. (looking into this).


cdjackson:
Thanks @t-8ch . I've not looked over this yet as the changes are extensive. I assume that the real changes are only to the POM? Is it possible to leave the existing code included in the repo for now, or does this not work for some reason?

cdjackson:
I also note that the build seems to be failing (according the to LGTM CI anyway)...

t-8ch:
You should be able to ignore the first commit completely.
I will take a look at the build.

t-8ch:
Ah, LGTM uses gradle. Is there an inherent reason for both build systems?

cdjackson:
Some systems needed to use gradle (some users needed it for Android).

t-8ch:
Hm, shouldn't users only consume the .jar file?
It shouldn't matter how it is generated.

cdjackson:
As said, this was required in some systems.

t-8ch:
I will see how it can support both

PR #533, opened 2019-02-13
532 Disable join not working properly

milansamardzic:
Regarding this comment, disable joining was not working properly.

`/**
     * Enables or disables devices to join the network.
     * <p>
     * Devices can only join the network when joining is enabled. It is not advised to leave joining enabled permanently
     * since it allows devices to join the network without the installer knowing.
     *
     * @param destination the {@link ZigBeeEndpointAddress} to send the join request to
     * @param duration sets the duration of the join enable. Setting this to 0 disables joining. As per ZigBee 3, a
     *            value of 255 is not permitted and will be ignored.
     * @return {@link ZigBeeStatus} with the status of function
     */`

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -1034,7 +1034,7 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
      * @return {@link ZigBeeStatus} with the status of function
      */
     public ZigBeeStatus permitJoin(final ZigBeeEndpointAddress destination, final int duration) {
-        if (duration < 0 || duration >= 255) {
+        if (duration <= 0 || duration >= 255) {
             logger.debug("Permit join to {} invalid period of {} seconds.", destination, duration);
             return ZigBeeStatus.INVALID_ARGUMENTS;
         }


cdjackson:
Thanks.

It seems that all you have changed is to prevent the value being set to 0. This should be an acceptable value (which would disable join) so I'm not sure why this change is needed? It should in fact stop the disable join from working shouldn't it as you have now made 0 (ie disable) an invalid argument?

cdjackson:
I note that the tests are also now failing as 0 is now not allowed.

milansamardzic:
@cdjackson thnx for comment. 
Yes, that is change. So per comment, if you want to disable join you need to pass 0 as a parameter, and there is a condition 
`
// pseudo code
if (duration < 0 || duration >= 255) { 
 return error;
} else {
return success;
}

So regarding to this, even you pass 0 as duration method will return success.
Maby to handle 0 as a separate case.. 


`

cdjackson:
I'm not sure that you understand this (or I misunderstand your issue).

A value of 0 is perfectly valid. Your change makes 0 an invalid argument. Please see below from the standards -:

image

The standard states -:

> If the PermitDuration parameter is not
equal to zero or 0xFF, the parameter is a number of seconds and joining is
permitted until it counts down to zero, after which time, joining is not permitted.
If the PermitDuration is set to zero, joining is not permitted.


By setting the parameter to 0, joining is disabled. I believe that this is correctly handled in the code - or at least your change is incorrect as it prevents 0 being sent.

milansamardzic:
Maby then the issue is on something else. While I testing after setting 0, joining is not disabled. It was still able to join...

cdjackson:
Ok, I'll close this and I suggest that you add more information to the issue you raised.

PR #536, opened 2019-02-15
Add some await statements to the ZclOtaUpgradeServerTest

### LINKED ISSUE #535: Flaky OTA Upgrade tests
The test `ZclOtaUpgradeServerTest.cancelUpgrade` sometimes (though rather rarely) fails.

The reason appears to be a race condition when checking for OTA status updates; consider, for instance, the following two lines from https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/305084f9a156658e51e30aa9a53943e6c6ca22c5/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/app/otaupgrade/ZclOtaUpgradeServerTest.java#L168:

```
server.setFirmware(otaFile);
assertTrue(otaStatusCapture.contains(ZigBeeOtaServerStatus.OTA_WAITING));
```

Calling `server.setFirmware` will result in a status update to `OTA_WAITING`; however, this update is propagated asynchronously to the `ZigBeeOtaStatusCallback`s (including the test class); in some cases, the update has not been propagated when checking for it.

I will try to provide a fix (not immediately, but in the next time).

hsudbrock:
The goal is to avoid race conditions that make the cancelUpdate test fail.

This resolves #535 .

### DIFF
--- a/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/app/otaupgrade/ZclOtaUpgradeServerTest.java
+++ b/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/app/otaupgrade/ZclOtaUpgradeServerTest.java
@@ -7,8 +7,9 @@
  */
 package com.zsmartsystems.zigbee.app.otaupgrade;
 
-import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks @hsudbrock

PR #540, opened 2019-02-18
Console: Add an option to remove a node after unsuccessful leave attempt

### LINKED ISSUE #537: Add possibility to remove a node that does not respond to Leave requests
This issue relates to https://github.com/openhab/org.openhab.binding.zigbee/issues/190, where the idea is to add some option to "force remove" an openHAB thing for which the ZigBee node does not respond to leave requests.

I think that it might not be straightforward to implement this in openHAB (as the ESH 'force leave' method will simply remove the thing from the thing registry, without giving the thing handler the chance to do some force-removal from its internal data structures).

However, I think that it would already be good to have that functionality in this library, including a console command for "force-removing" a node.

My suggestion would be to add a method `leave(final Integer destinationAddress, final IeeeAddress leaveAddress, boolean forceRemove)` to the `ZigBeeNetworkManager`; for `forceRemove==true` it would remove the node from the list of nodes, even if the node does not respond to the `Leave` command. The existing `leave(destinationAddress, leaveAddress)` would then simply call `leave(destinationAddress, leaveAddress, false)`.

What do you think, @cdjackson ? If you don't have immediate concerns, I could create a PR for this change.

hsudbrock:
Resolves #537 

As discussed in #537, I have added an option to the `Leave` CLI command that will first try to make a node leave the network, and, if the node does not respond to the leave request, removes the node from the list of nodes in the network manager.

I tried to implement this by first calling `networkManager.leave`, and , if this fails, calling `networkManager.removeNode`. But I did not see a way to find out whether the `leave` call succeeded, except setting a timeout and checking whether the node is still there after that timeout.

Hence, I have added a version of the `networkManager.leave` command that will remove the node if there is no successful response to the leave command.

Does that sound ok for you, Chris?

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -1091,13 +1091,27 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
     }
 
     /**
-     * Sends a ZDO Leave Request to a device requesting that an end device leave the network.
+     * Sends a ZDO Leave Request to a device requesting that an end device leave the network. 
+     * Only in case of a successful response is the corresponding node removed from the node list.
      *
      * @param destinationAddress the network address to send the request to - this is the device parent or the the
      *            device we want to leave.
      * @param leaveAddress the {@link IeeeAddress} of the end device we want to leave the network
      */
     public void leave(final Integer destinationAddress, final IeeeAddress leaveAddress) {
+        leave(destinationAddress, leaveAddress, false);
+    }
+    
+    /**
+     * Sends a ZDO Leave Request to a device requesting that an end device leave the network. 
+     *
+     * @param destinationAddress the network address to send the request to - this is the device parent or the the
+     *            device we want to leave.
+     * @param leaveAddress the {@link IeeeAddress} of the end device we want to leave the network
+     * @param forceNodeRemoval If this flag is set, then the corresponding  node is removed from the node list 
+     *            even if there was no success response to the leave command.
+     */
+    public void leave(final Integer destinationAddress, final IeeeAddress leaveAddress, boolean forceNodeRemoval) {
         final ManagementLeaveRequest command = new ManagementLeaveRequest();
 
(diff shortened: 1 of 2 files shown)


hsudbrock:
I am not quite sure whether I understand your question regarding the structure of the code - which part of the code are you referring to?

cdjackson:
The bit where I put the comment ;)

hsudbrock:
Could it be that you still need to submit the review with the comment? I can only see the following comment, which appears not to be attached to a line in the code:

> Thanks Henning - I'm ok with the general solution - only one question about the structure of the code. Maybe you did it this way so that the logging is specific to the instances, so if you think this is best, then I'm probably happy with this...

cdjackson:
Hmmm - strange - I'll add it again.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -1122,6 +1136,14 @@ public void run() {
                     } else {
                         logger.debug("{}: No successful response received to leave command (status code {})",
                                 leaveAddress, response.getStatusCode());
+                        if (forceNodeRemoval) {
+                            ZigBeeNode node = getNode(leaveAddress);
```
cdjackson: I was wondering if this could be consolidated iwth the above check - something like -:

```
                    if (response.getStatusCode() == 0 || forceNodeRemoval) {
```

This would just avoid the (admittedly relatively minor) duplication as in both cases we call ```removeNode()```

hsudbrock: I see it now, thanks for adding the comment again.

I actually did not include the `forceNodeRemoval` case in the `response.getStatusCode() == 0` case, so that one can log explicitly that the node was "force-removed".

Looking at the code once more now, I think that one could maybe consolidate by pulling the `node = getNode(leaveAddress)` out of the `if` block and put it directly after the `sendTransaction(...,...).get()`, so that one does not have to get the node twice - but let me take a look at this tomorrow when I am at a bigger screen.

hsudbrock:
I have pushed a commit that consolidates the different cases after attempting to make a device leave the network.

cdjackson:
Thanks Henning - I think that's more readable.

PR #541, opened 2019-02-19
Add manufacturer and model information to the nodelist command

hsudbrock:
This PR is an attempt to improve the `nodelist` CLI command, so that it provides manufacturer and model ID from the basic cluster on the first endpoint, if available.

The reason for adding this is that I am quite often in the situation that I want to get information about or interact with a device using the console, for which I need the endpoint address (e.g., `1234/1`) - which I usually do not know by heart. I then call `nodelist`, which gives me a list of 10 or so nodes, and then I have to manually issue `endpoint <nodeaddress>/1` commands and check manufacturer/model in the basic cluster attributes until I found the correct node address. I found this somewhat tedious.

With this change to the `nodelist` command, the command attempts to read manufacturer and model ID from the basic cluster of the first endpoint; to make the output a little easier to read, I have also added bars as table line/column separators. The output looks like in the following example:

```
Network | Addr |      IEEE Address |  Logical Type |  EP |                   Profile |     Device Type |    Manufacturer |           Model
------------------------------------------------------------------------------------------------------------------------------------------
      0 | 0000 |  000D6F000DCFA37F |   COORDINATOR |   1 |    ZIGBEE_HOME_AUTOMATION |   ON_OFF_SWITCH |                 |                
------------------------------------------------------------------------------------------------------------------------------------------
  29750 | 7436 |  00124B0006233B7E |    END_DEVICE |   1 |    ZIGBEE_HOME_AUTOMATION |   ON_OFF_SWITCH |     Bitron Home |       902010/23
------------------------------------------------------------------------------------------------------------------------------------------
  40309 | 9D75 |  000D6F000E36AAE0 |        ROUTER |   1 |    ZIGBEE_HOME_AUTOMATION |   ON_OFF_SWITCH |      CentraLite |         3200-de
------------------------------------------------------------------------------------------------------------------------------------------
  60722 | ED32 |  00124B00028A1B7D |        ROUTER |   1 |    ZIGBEE_HOME_AUTOMATION |   ON_OFF_SWITCH |     Bitron Home |       902010/28
        |      |                   |               |   4 |    ZIGBEE_HOME_AUTOMATION |   ON_OFF_SWITCH |                 |                
------------------------------------------------------------------------------------------------------------------------------------------
```

What do you think about this, Chris? I am aware that adding infos from the basic cluster to the output of the `nodelist` command kind of breaks the barrier between the `ZigBee` and the `ZCL` layers - but I think that having this info in there would be really useful to console users.

### DIFF
--- a/com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleNodeListCommand.java
+++ b/com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleNodeListCommand.java
@@ -59,21 +59,19 @@ public class ZigBeeConsoleNodeListCommand extends ZigBeeConsoleAbstractCommand {
         }
 
         Collections.sort(nodeIds);
-        String tableHeader = String.format("%7s | %4s | %17s | %13s | %3s | %25s | %15s | %15s | %15s", "Network",
-                "Addr", "IEEE Address", "Logical Type", "EP", "Profile", "Device Type", "Manufacturer", "Model");
+        String tableHeader = String.format("%7s  %4s  %17s  %13s  %3s  %25s  %15s  %15s  %15s", "Network", "Addr",
+                "IEEE Address", "Logical Type", "EP", "Profile", "Device Type", "Manufacturer", "Model");
         out.println(tableHeader);
-        out.println(createPadding(tableHeader.length(), '-'));
 
         for (Integer nodeId : nodeIds) {
             printNode(networkManager.getNode(nodeId), out);
-            out.println(createPadding(tableHeader.length(), '-'));
         }
     }
 
     private void printNode(ZigBeeNode node, PrintStream out) {
-        String nodeInfo = String.format("%7d | %04X | %17s | %13s |", node.getNetworkAddress(),
-                node.getNetworkAddress(), node.getIeeeAddress(), node.getLogicalType());
-        String nodeInfoPadding = String.format("%7s | %4s | %17s | %13s |", "", "", "", "");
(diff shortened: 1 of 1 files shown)


cdjackson:
I'm fine with this in general. I'm not really sure it needs the | separators (or even the horizontal separators) and we don't have them in other such tables. Maybe the horizontal lines are ok here since we have the endpoints so the first few columns may be empty....

I guess it's probably personal preference - unless you really think it makes a major difference to readability, my preference would be to remove them simply to keep the formatting more common with other tables (eg security keys comes to mind).

WDYT?

hsudbrock:
I will remove the tables lines. (I don't have a strong opinion on the table lines - I think that it is more readable, but I also see your point that it is not in line with the other commands. I guess you are right that it is personal preference.)

hsudbrock:
Just for info, the output now looks as follows (same devices as in comment above):

```
Network  Addr       IEEE Address   Logical Type   EP                    Profile      Device Type     Manufacturer            Model
      0  0000   000D6F000DCFA37F    COORDINATOR    1     ZIGBEE_HOME_AUTOMATION    ON_OFF_SWITCH                                  
  29750  7436   00124B0006233B7E     END_DEVICE    1     ZIGBEE_HOME_AUTOMATION    ON_OFF_SWITCH      Bitron Home        902010/23
  40309  9D75   000D6F000E36AAE0         ROUTER    1     ZIGBEE_HOME_AUTOMATION    ON_OFF_SWITCH       CentraLite          3200-de
  60722  ED32   00124B00028A1B7D         ROUTER    1     ZIGBEE_HOME_AUTOMATION    ON_OFF_SWITCH      Bitron Home        902010/28
                                                   4     ZIGBEE_HOME_AUTOMATION    ON_OFF_SWITCH                                  
```

cdjackson:
Thanks Henning.

I have toyed with putting the vertical lines in in the past, but in some cases it makes the tables very wide, then they wrap, and it's a totally unreadable mess.

PR #552, opened 2019-02-28
Permit querying multiple attributes via CLI read command

hsudbrock:
This exploits the newly added support for reading multiple attributes with a `Read Attribute` command.


cdjackson:
Nice - thanks Henning.

PR #555, opened 2019-03-01
CLI nodelist: Do not force read of modelId/manufacturer

### LINKED ISSUE #553: Nodes cli command issues getting basic information
Following #541 if there are issues getting the basic information, then the nodes cli command gets spread over a long duration making it difficult to use now when we just want to see the list of nodes. I would suggest that the BASIC information that was added in this change should not be requested by default, and we should have an option to request it.

This could be implemented with a Long.MAX_VALUE as the refresh time to allow the information to be displayed if it is already known, but not requested if it isn't.

@hsudbrock WDYT?

hsudbrock:
This resolves #553.

Unlike what I wrote in #553, I did not introduce a flag for the command; instead, the command now displays the last value of the model ID and the manufacturer (unless these values were never read from the device, in which no model ID / manufacturer is displayed in the node list for the device).

If one wants to have model ID and manufacturer in the nodelist output, and they have not yet been read, one simply has to execute a `read <endpoint> 0 4 5` command to read both attributes once.

### DIFF
--- a/com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleNodeListCommand.java
+++ b/com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleNodeListCommand.java
@@ -7,6 +7,8 @@
  */
 package com.zsmartsystems.zigbee.console;
 
+import static com.zsmartsystems.zigbee.zcl.clusters.ZclBasicCluster.*;
+
 import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks Henning.

Regarding this point -:

> If one wants to have model ID and manufacturer in the nodelist output, and they have not yet been read, one simply has to execute a `read <endpoint> 0 4 5` command to read both attributes once.

Just to note that there is also the ```info``` command that will read all these attributes as well...

hsudbrock:
>  Just to note that there is also the `info` command that will read all these attributes as well..
Ah, that's even better, thanks Chris!

cdjackson:
I'm not sure if it's better - but it's another option ;)

PR #556, opened 2019-03-01
Support multiple attribute updates per report

t-8ch:
# Description

As the listeners are executed asynchronously all parts of the report
will have been processed when the listeners are called.
Attribute.getLastValue() will be the same for all listeners.

# Usecase

I want to add support for a Xiami Mini Wireless switch to the openhab binding.
On each press the switch sends a single attributereport command with two reports in it.
One for pressed, one for released. Unfortunately the attribute listener inside the binding can't match on the different values, as it only receives a reference to the attribute and has to fetch the last value out of it, which is always the last one.

# Comments

Currently this changes the AttributeListener interface.
If it is preferred to be fully backwards compatible we could add `default` variants for both interface methods and new users of the interface could use the new method.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
@@ -826,13 +826,14 @@ public abstract class ZclCluster {
      * Notify attribute listeners of an updated {@link ZclAttribute}.
      *
      * @param attribute the {@link ZclAttribute} to notify
+     * @param value the current value of the attribute
      */
-    private void notifyAttributeListener(final ZclAttribute attribute) {
+    private void notifyAttributeListener(final ZclAttribute attribute, final Object value) {
         for (final ZclAttributeListener listener : attributeListeners) {
             NotificationService.execute(new Runnable() {
                 @Override
                 public void run() {
-                    listener.attributeUpdated(attribute);
+                    listener.attributeUpdated(attribute, value);
                 }
             });
         }
@@ -912,8 +913,9 @@ public abstract class ZclCluster {
             logger.debug("{}: Unknown attribute {} in cluster {}", zigbeeEndpoint.getEndpointAddress(), attributeId,
                     clusterId);
         } else {
-            attribute.updateValue(normalizer.normalizeZclData(attribute.getDataType(), attributeValue));
-            notifyAttributeListener(attribute);
+            Object value = normalizer.normalizeZclData(attribute.getDataType(), attributeValue);
+            attribute.updateValue(value);
+            notifyAttributeListener(attribute, value);
         }
     }
 
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclAttributeListener.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclAttributeListener.java
@@ -18,6 +18,7 @@ public interface ZclAttributeListener {
      * Called when an attribute is updated
      *
      * @param attribute the {@link ZclAttribute} that has been updated
+     * @param value the new value of the attribute
      */
-    void attributeUpdated(ZclAttribute attribute);
+    void attributeUpdated(ZclAttribute attribute, Object value);
 }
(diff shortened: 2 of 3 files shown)


cdjackson:
While this makes some sense, it is obviously a breaking change. I'll take a look, but if merged it wouldn't be into master, but a later version.

t-8ch:
I will add a second commit wich preserves both the API and semantics for current users tomorrow.

cdjackson:
I’m not sure that really helps - it would still be a breaking change as the new method would need to be implemented. It might just be worth biting the bullet and breaking it…

Feel free to provide the second method, and we can discuss further...

t-8ch:
It would be a default method in the interface. (Given we can rely on Java 8)

cdjackson:
Unfortunately at the moment we can’t use all Java 8 features in the main code (ie it can be used for tests) due to some Android users. I will check if this can be used (from a quick look, it might be ok and not dependant on specific android versions, but I will check.

t-8ch:
Ah ok. Thanks!
Another possibility would be to create a second interface with the other method, overload  `addListener()` to accept either and wrap the old listeners into an adapter for the new interface.

t-8ch:
The [android docs](https://developer.android.com/studio/write/java8-support) say default methods are supported on all API levels.
I assume it would still be possible to try to build the plugin with an old toolchain that does not yet support the feature and run into errors.

t-8ch:
@cdjackson For the same device I would also like to support the double clicks.
Unfortunately there the `ReportAttributesCommand` specifies an attribute that is not part of the known attributes (missing in xml and attribute map).
What do you think about another overload, that gets called with the raw `int attributeId` and value?

cdjackson:
I’m not super keen in this - it means both methods need to be called, and I think things start getting messy.  In the near future it should be possible to add new attributes to the cluster that are manufacturer specific, so this ought to allow this to be solved in that way.  Is that ok?

t-8ch:
Yep, understandable. It would be fine for my usecase.

It would would not support (probably very broken) devices that even sends non-deterministic attributes. But I don't know if those existed and should be supported.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -824,15 +824,12 @@ public void removeAttributeListener(final ZclAttributeListener listener) {
...
-    private void notifyAttributeListener(final ZclAttribute attribute) {
+    private void notifyAttributeListener(final ZclAttribute attribute, Object value) {
         for (final ZclAttributeListener listener : attributeListeners) {
-            NotificationService.execute(new Runnable() {
-                @Override
-                public void run() {
-                    listener.attributeUpdated(attribute);
-                }
-            });
+            NotificationService.execute(() -> listener.attributeUpdated(attribute, value));
```
cdjackson: Can you change this back to use the non-streaming API please.

t-8ch: Hm, these aren't streams. It is a lambda, which should be supported according to the docs. I can remove them though.

cdjackson:
As far as I know this was introduced in Android 24 - the stated minimum requirement is API 19. Maybe you have better information? Can you forward the link to the docs you used please?

t-8ch:
https://developer.android.com/studio/write/java8-support

Streams are indeed annotated with apilevel 24, but lambdas should work everywhere

cdjackson:
I’ll check - I read that Lambdas were introduced at Android 7, but again, I might have misread.

cdjackson:
Looks like it is toolchain dependant rather than SDK dependant… I’ll therefore need to check further with other companies - it may take a little time so we might skip this for now - sorry.

cdjackson:
The other option is to change it - which might be quicker. I leave it up to you though.

t-8ch:
The PR also depends on default interface methods that have the same version info as the lambdas. Are you sure the default methods work?

cdjackson:
I have not tried this with Android personally, so no - I’m not sure.

Maybe the easiest (well, safest!) thing then is remove the default as well? There are a number of breaking changes in 1.2 around this area, so maybe it’s cleaner to avoid using the default method from the start rather than add it and then deprecate it in 1.3?

t-8ch:
The alternatives would be
* Change the existing interface methods and force users to adapt their code
* (Add a new method to the interface and force users to implement it via a noop, does not really make sense)
* Introduce a new interface with the new method.
* What we discussed before.

If we can get away with breaking library user code, then let's just change the method. Ok?

cdjackson:
Yep - as I said, we already have a number of breaking changes, so I think the cleanest is to change the interface and require users to update code. It’s not super nice, but I think it’s cleanest and safest.

t-8ch:
Thanks. I misinterpreted that sentence. I will update the PR.

cdjackson:
No probs - thanks.

t-8ch:
@cdjackson Done

cdjackson:
Thanks.

PR #557, opened 2019-03-02
Introduce ZigBeeNetworkState enum to separate transport and network states

### LINKED ISSUE #548: Add possibility to recognize if a network initialization/creation is in progress
As discussed in #518 we want to avoid multiple initialization of the network. Therefore we propose to introduce a possibility to check if an initialization is already in progress.

The idea is to introduce a flag `initializing` in the network manager.
- The flag must be set to `true` when `ZigBeeNetworkManager.initialize()` and/or `ZigBeeNetworkManager.startup(boolean)` are called and set back to `false` when these methods are done.
- There must be a public method to check the flag

We want to use this flag in the binding in order to skip any other initialization.

### LINKED ISSUE #554: Improve handling of ZigBeeNetworkState
It is proposed that the ```ZigBeeNetworkStateListener``` should use a ```ZigBeeNetworkState``` enumeration rather than the current transport enum. This should avoid confusion by ensuring a clear separation of the transport and network layers.

In addition, updates from the transport layer to ```OFFLINE``` should be filtered during initialisation so that the ```ZigBeeNetworkState``` does not change state while initialisation is in progress. If initialisation fails (eg at the completion of the ```ZigBeeNetworkManager.startup()``` call, then a transition to ```OFFLINE``` is still permitted.

cdjackson:
This separates the state of the transport layer from the network state. It introduces a new ```ZigBeeNetworkState``` enum, and the ```ZigBeeNetworkStateListener``` now receives updates with this new enum and not the ```ZigBeeTransportState```.

When refactoring it is noted that state names remain the same between the two - just the class name changes.

Closes #548 
Closes #554

### DIFF
--- a/com.zsmartsystems.zigbee.console.main/src/main/java/com/zsmartsystems/zigbee/console/main/ZigBeeConsole.java
+++ b/com.zsmartsystems.zigbee.console.main/src/main/java/com/zsmartsystems/zigbee/console/main/ZigBeeConsole.java
@@ -35,6 +35,7 @@ import com.zsmartsystems.zigbee.ZigBeeEndpoint;
 import com.zsmartsystems.zigbee.ZigBeeGroupAddress;
 import com.zsmartsystems.zigbee.ZigBeeNetworkManager;
 import com.zsmartsystems.zigbee.ZigBeeNetworkNodeListener;
+import com.zsmartsystems.zigbee.ZigBeeNetworkState;
 import com.zsmartsystems.zigbee.ZigBeeNetworkStateListener;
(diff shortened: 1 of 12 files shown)


cdjackson:
This pull request **introduces 1 alert** when merging 970a6c387cf5342de7bf854dafe77442652f7266 into 65dbb3590cc6c397a3dca9bae249f625343c4488 - [view on LGTM.com](https://lgtm.com/projects/g/zsmartsystems/com.zsmartsystems.zigbee/rev/pr-4f04e9def3dc4da9817bf7b8bfb545ed0589215e)

**new alerts:**

* 1 for Inconsistent synchronization of getter and setter

---

PR #559, opened 2019-03-03
Transaction queue - Fix race condition in Telegesis management of SEQ/ACK/NACK

puzzle-star:
Make all responses from the Telegesis dongle be pipelined to be treated in order.

This prevents ACK responses from being processed by an independent thread before the handling of the SEQ response has updated the transaction IDs map.

This also ensures the command response from the remote device is not processed before the TX_ACK/NAK is processed

This comes from [this discussion at OH forum](https://community.openhab.org/t/zigbee-binding/15763/2050?u=puzzle-star)


cdjackson:
Thanks.

PR #560, opened 2019-03-03
Remove 'synchronized (this)' blocking for writing network state file in updateNode/addNode/removeNode

puzzle-star:
In updateNode / addNode / removeNode, access to nodeListeners list does not need to be synchronized as, the list is unmodifiable (and fully replaced by another one when listeners are added / removed)

Serializing afterwards needs to be synchronized, but it can be done in a separate thread, so this is called now through a separate thread from the executorService.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -1295,22 +1295,27 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
             networkNodes.remove(node.getIeeeAddress());
         }
 
-        synchronized (this) {
-            for (final ZigBeeNetworkNodeListener listener : nodeListeners) {
-                NotificationService.execute(new Runnable() {
-                    @Override
-                    public void run() {
-                        listener.nodeRemoved(node);
-                    }
-                });
-            }
-
-            node.shutdown();
-
-            if (networkStateSerializer != null) {
-                networkStateSerializer.serialize(this);
-            }
+        for (final ZigBeeNetworkNodeListener listener : nodeListeners) {
+            NotificationService.execute(new Runnable() {
+                @Override
+                public void run() {
+                    listener.nodeRemoved(node);
+                }
+            });
         }
+
+        node.shutdown();
+
+        executorService.execute(new Runnable() {
+            @Override
+            public void run() {
+                synchronized (ZigBeeNetworkManager.this) {
+                    if (networkStateSerializer != null) {
+                        networkStateSerializer.serialize(ZigBeeNetworkManager.this);
+                    }
+                }
+            }
+        });
     }
 
     /**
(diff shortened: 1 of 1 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -1295,22 +1295,27 @@ public void removeNode(final ZigBeeNode node) {
...
-            node.shutdown();
+        node.shutdown();
 
-            if (networkStateSerializer != null) {
-                networkStateSerializer.serialize(this);
+        executorService.execute(new Runnable() {
+            @Override
+            public void run() {
+                synchronized (ZigBeeNetworkManager.this) {
+                    if (networkStateSerializer != null) {
```
cdjackson: Wouldn't it be better to put this check outside of the thread?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -1335,24 +1340,29 @@ public void addNode(final ZigBeeNode node) {
...
+            });
+        }
 
-            if (networkStateSerializer != null) {
-                networkStateSerializer.serialize(this);
+        executorService.execute(new Runnable() {
+            @Override
+            public void run() {
+                synchronized (ZigBeeNetworkManager.this) {
+                    if (networkStateSerializer != null) {
```
cdjackson: Wouldn't it be better to put this check outside of the thread?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -1389,24 +1399,29 @@ public void updateNode(final ZigBeeNode node) {
...
```
cdjackson: Wouldn't it be better to put this check outside of the thread?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
puzzle-star: If it were outside, networkStateSerializer could be set to null by another thread between the check and calling the scheduled method, leading to a NPE.
Setting networkStateSerializer is done synchronized on (this).

puzzle-star: If it were outside, networkStateSerializer could be set to null by another thread between the check and calling the scheduled method, leading to a NPE.
Setting networkStateSerializer is done synchronized on (this).

puzzle-star: If it were outside, networkStateSerializer could be set to null by another thread between the check and calling the scheduled method, leading to a NPE.
Setting networkStateSerializer is done synchronized on (this).

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
puzzle-star: Same as above ...

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
puzzle-star: Same as above...

cdjackson:
Fair enogh - do you plan to do an update?

puzzle-star:
I am not at home now. I can move the duplication to a separate method later (i.e triggerNetworkSerialization), but it will have to be changed for the datastore, and maybe the three calls will be different then, do maybe it is not worth it...

Let me know what you prefer

Pedro

cdjackson:
Ok, let’s leave it as is. I hope to release 1.2 this month anyway.

PR #561, opened 2019-03-03
Remove global dependency on logging implementation

t-8ch:
Currently the library pulls in a dependency on log4j and slf4j-log4j12
although they are not used.
Also libraries should not pull in logging implementations as it will
conflict with other logging implementations on the classpath.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,11 @@ allprojects {
  version = '1.1.10-SNAPSHOT'
 }
 
+ext {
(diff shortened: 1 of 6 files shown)


cdjackson:
Thanks. Does this still log to console during tests?

t-8ch:
It should (at least for the main library, as the other ones don't have explicit log4j configs, I left it out).

cdjackson:
Ok, thanks, then looks fine to me.

PR #580, opened 2019-03-19
Do not set link state to true after SPI connect

### LINKED ISSUE #578: For Ember/SPI the network state is set to ONLINE too early
When using an Ember chip via SPI, the link state is set to `true` when the response to the `SPI Status` command has the status bit set (see [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java#L372)). 

The issue is that this results in the network state being set to `ONLINE` (see [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java#L735)), although the network address in the `ZigBeeDongleEzsp` is not yet set. (When trying to retrieve the network address at this early point [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java#L729), the chip cannot yet provide it and replies with `0xFFFE`.)

The consequence of this issue is an NPE [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java#L954), where the code now tries to unbox `null` into an `int` variable. This NPE prevents the setup that should be done when getting `ONLINE` from happening (i.e., adding the local node, starting up extensions, notifying listeners).

When the network state is later set to `ONLINE` again [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java#L513), the network key is available. But the code needed to be run when getting `ONLINE` is not executed, because the state is already online (see [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java#L929)). I.e., extensions are never started, and listeners are not notified that the state is `ONLINE`.

Why did this issue not occur earlier? The issue got visible only by #550; this PR moved setting the network state to `INITIALIZING` before the initialization of the transport ([here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java#L321)); in consequence, with this PR the 'too early' setting to `ONLINE` became a valid status transition; before this PR it was an invalid status transition because the network state was still `UNINITIALIZED`, so it was simply skipped.

I am not completely sure what the correct fix for this issue is. I have the impression that the status should not be set to `ONLINE` just because `SPI` is ready (as the network key is not ready then). So should the update of the link state to `true` [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/03bdd72f46762a7602f39ec81ac1a61c013bcf22/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java#L372) just be removed? Or would this have other bad consequences, that I just don't see? I would be happy about your feedback, @cdjackson .

hsudbrock:
This resolves #578 

Currently, for SPI the link state is set to `true` if (a) a SPI Status command with status bit set is received, or if (b) an `EzspStackStatusHandler` callback with status `EMBER_NETWORK_UP` is received. (For ASH, only (b) is the case.)

However, if the library handles a link state change to `true`, it assumes that it can read the network address from the chip. This is, however, not possible if only the SPI connection is ready, but the network is not yet up on the chip.

Hence, the link status should only be set to `true` if the `EMBER_NETWORK_UP` status is received.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java
@@ -368,9 +368,7 @@ public class SpiFrameHandler implements EzspProtocolHandler {
                         spiTransactionComplete();
                         stateConnected = (packetData[0] & SPI_MASK_READY) == SPI_MASK_READY;
                         logger.debug("SPI Protocol Ready: {}", stateConnected);
-                        if (stateConnected) {
-                            frameHandler.handleLinkStateChange(true);
-                        } else {
+                        if (!stateConnected) {
                             outputFrame(requestSpiStatus);
                         }
                         break;
--- a/com.zsmartsystems.zigbee.dongle.ember/src/test/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandlerTest.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/test/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandlerTest.java
@@ -7,10 +7,7 @@
  */
 package com.zsmartsystems.zigbee.dongle.ember.internal.spi;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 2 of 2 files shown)


cdjackson:
Agreed - thanks Henning.

Issue #583, opened 2019-03-29
Network re-initialization still sets the network state to OFFLINE


TomTravaglino:
This issue is related to #548 #554 #557 and to [this one in the binding](https://github.com/openhab/org.openhab.binding.zigbee/issues/398)

In the issues mentioned above we discussed the problem that the reconnect job in the binding becomes active when the initialization is in progress. A solution was provided in #554 which prevents the ZigBeeNetworkState from going OFFLINE during initialization.
Despite this solution I'm still experiencing the network state going OFFLINE during initialization. I investigated the issue and could reproduce how it happens.

- My app calls `ZigBeeNetworkManager.startup(true)`.
- `ZigBeeNetworkManager.startup` calls `transport.startup(reinitialize)`. In my case `transport` is an instance of `ZigBeeDongleEzsp`.
- `ZigBeeDongleEzsp.startup` calls `EmberNetworkInitialisation.formNetwork(...)`.
- `EmberNetworkInitialisation.formNetwork` calls `EmberNcp.leaveNetwork()`.
- `EmberNcp.leaveNetwork` sends a `EzspLeaveNetworkRequest`.
- By sending `EzspLeaveNetworkRequest` a `EzspStackStatusHandler [status=EMBER_NETWORK_DOWN]` is generated.
- `ZigBeeDongleEzsp.handlePacket(EzspFrame)` receives `EzspStackStatusHandler` and calls `ZigBeeDongleEzsp.handleLinkStateChange(false)`.
- `ZigBeeDongleEzsp.handleLinkStateChange` calls `ZigBeeNetworkManager.setTransportState(ZigBeeTransportState)` where the network state changes from ONLINE to OFFLINE and this makes the reconnect job run.

Maybe the network state should be set to INITIALISING at an earlier point (before forming the network) so that the state change to OFFLINE can be filtered as implemented in fix #554.

cdjackson:
Please define what version of the libraries you are using?

TomTravaglino:
We're testing zss 1.2.0 which contains the fix #554.

Regarding my statement
>Maybe the network state should be set to INITIALISING at an earlier point ...

Where do you think is the right place to set the network state to INITIALISING?

cdjackson:
> Where do you think is the right place to set the network state to INITIALISING?

[...]

PR #590, opened 2019-04-13
Network state transition update

cdjackson:
Update ```ZigBeeNetworkState``` transitions as per #583 discussion.

Adds a ```reinitialize``` method to ```ZigBeeNetworkManager``` to take the network offline and allow reconfiguration.


cdjackson:
@triller-telekom this is updated with the ```reinitialize``` method - are you able to easily test to see if it resolves your issue before I merge?

triller-telekom:
I can probably give it a try tomorrow

triller-telekom:
One question though:

```
    public ZigBeeStatus reinitialize() {
        // Set the state to INITIALISING before reconfiguring the transport layer
        // to ensure we don't pass the OFFLINE state to the application
        setNetworkState(ZigBeeNetworkState.INITIALISING);

         return ZigBeeStatus.SUCCESS;
    }
```

This seems to only set the state to INITIALISING. Does that mean that my application has to call `reinitialize()` first and then `startup(true)` to make the network manager write the new configuration values onto the chip?

I thought your `reinitialize()` method would take care of that internally.

Or did you do it to have it consistent with `initialize()` +`startup()`, where also both methods have to be called by the application?

cdjackson:
Yes - the reinitialise method is to put the state back into INITIALIZE - so the same state as if you had called the initialise() method. It doesn’t actually initialise anything, but as some methods can only be called between initialise and startup, you need to put the system back into INITIALIZE state, update the configuration, then call startup().

There is still a potential issue with this, and that is that there’s currently no way to take the transport OFFLINE which may also be needed in some instances. This requires a further update to the transport interface.

triller-telekom:
I can confirm that it works.

I have re-configured the network via

```
networkManager.setZigBeePanId(Integer.parseInt(panID));
networkManager.setZigBeeExtendedPanId(new ExtendedPanId(extendedPanID));
networkManager.setZigBeeChannel(ZigBeeChannel.create(Integer.parseInt(channel)));
networkManager.setZigBeeLinkKey(new ZigBeeKey(linkKey));

ZigBeeKey key = new ZigBeeKey(networkKey);
key.setOutgoingFrameCounter(frameCounter);
networkManager.setZigBeeNetworkKey(key);
```
and then did this:

```
networkManager.reinitialize();
networkManager.startup(true);
```

Afterwards the network is back working fine. Also the `ZigbeeNetworkState` does **NOT** go to OFFLINE in-between, which is what we want.

cdjackson:
Thanks for testing @triller-telekom

PR #612, opened 2019-04-29
ZigbeeNetworkManager: Possibility to add node if it doesn't exist

triller-telekom:
This new method provides the possibilty to add a node to the network if
it does not exist yet, without updating an existing node. Thus, if the node
exists, it will simply be not added again.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -1294,6 +1294,39 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
             return;
         }
 
+        notifyListenersAboutAddedNode(node);
+    }
+
+    /**
+     * Adds a {@link ZigBeeNode} to the network if it doesn't exist yet
+     *
+     * @param node the {@link ZigBeeNode} to add
+     */
+    public void addNodeIfNotExist(final ZigBeeNode node) {
+        if (node == null) {
+            return;
+        }
+
+        synchronized (networkNodes) {
+            // Don't add if the node is already known
+            // We especially don't want to notify listeners
+            if (networkNodes.containsKey(node.getIeeeAddress())) {
+                logger.debug("{}: Tried to add Node {} to the network, but it already exists", node.getIeeeAddress(),
+                        node.getNetworkAddress());
+                return;
+            }
+            networkNodes.put(node.getIeeeAddress(), node);
+            logger.debug("{}: Node {} added to the network", node.getIeeeAddress(), node.getNetworkAddress());
+        }
+
+        if (networkState != ZigBeeTransportState.ONLINE) {
+            return;
+        }
+
+        notifyListenersAboutAddedNode(node);
+    }
+
+    private void notifyListenersAboutAddedNode(final ZigBeeNode node) {
         for (final ZigBeeNetworkNodeListener listener : nodeListeners) {
             NotificationService.execute(new Runnable() {
                 @Override


triller-telekom:
As briefly mentioned already, a possibility to add node to the networkmanager, only if it doesn't exist yet, would be nice.

The use case would be a transfer of existing zigbee nodes to a new network, i.e. if one replaces one zigbee stick with another and wants to replay a backup with all his devices.

The new method can be used to populate the networkmanager with known nodes, which might currently be sleeping and thus would not necessarily appear in the new network immediately, but we know they are there, because they were paired before.

Since the dev-1.2.0 branch is collecting API breaking changes and I extend the API, I filled the PR against this branch.

cdjackson:
If you are reloading the network configuration when transitioning to a new coordinator, why not just migrate the complete network while the network is offline. Wouldn't that be safer, and also more logical?

triller-telekom:
> just migrate the complete network while the network is offline. Wouldn't that be safer, and also more logical?

I agree, but I wonder how to achieve that:

In order to add or remove nodes, I need access to the `ZigbeeNetworkManager`, which holds all the nodes. So this has to be accessible. For now it is in the ONLINE state, which is why I propose to add this new method.

Which state do you suggest for the `ZigbeeNetworkManager`, so I can safely add new nodes? You say "OFFLINE", so shall I call `ZigbeeNetworkManager.shutdown()`?

If I remember correctly you said that the `ZigbeeNetworkManager` cannot recover from the SHUTDOWN state and instead one should instantiate a new one. However I cannot easily create a new instance.

One idea would be to patch the `addNode()` and `updateNode()` methods to reject any add requests if the network is still in INITIALISING state. This way I could call `ZigbeeNetworkManager#reinitialize()` first, and then add my nodes, before I can call `ZigbeeNetworkManager.startup()` again. @cdjackson  WDYT?

Background: The openHAB `ZigbeeCoordinatorHandler` holds the `ZigbeeNetworkManager` and I have the coordinator up and running in order to be able to "talk to the chipset via ZSS".

cdjackson:
If you are changing network coordinators, why can't these devices be added during the change?
- ie take the old coordinator offline, create a new coordinator, and add all the existing nodes, then bring the new coordinator online. If you are changing coordinators, then there are other things that need to be completed at the same time such as setting all the network counters, keys etc.

> Which state do you suggest for the ZigbeeNetworkManager, so I can safely add new nodes?

I think ```INITIALIZING``` is where it is currently done. All nodes are added to the system from the store at this stage and no notifications are sent out to listeners until the system comes ```ONLINE```.

> If I remember correctly you said that the ZigbeeNetworkManager cannot recover from the SHUTDOWN state 

Correct.

> One idea would be to patch the addNode() and updateNode() methods to reject any add requests if the network is still in INITIALISING state.

We can't reject add requests during ```INITIALIZING``` as I believe that's where they are currently added....

I just checked the current code and it changed slightly with the new database implementation - it now does the database restore in ```UNINITIALIZED``` but the same points above are otherwise correct in that there should be no updates to the listeners until the network is ```ONLINE```. Additionally, if the network is not ```ONLINE``` other parts of the system (eg discovery) should not be doing any updates to the nodes and I think this was one of you concerns that there could be a clash.

So, I think it should be fine to add nodes in the ```INITIALIZING``` state. We could easily move the state in which the nodes are added from the database from ```UNINITIALIZED``` to ```INITIALIZING``` by moving this down one line of code (ie at the moment the database is restored then the state updated to ```INITIALIZING``` - we could just swap the order if this is important.

Does this work for you?

One other point - since you are using the new database system, you could easily be keeping the same network by copying the database files to the new coordinator folder (I assume you are using the database implementation I wrote for OH?).

triller-telekom:
So you are recommending that if the `ZigbeeNetworkManager` is set into the `INITIALIZING` state via `reinitialize()`, I could safely manipulate the nodes, since you say:

> Additionally, if the network is not ONLINE other parts of the system (eg discovery) should not be doing any updates to the nodes and I think this was one of you concerns that there could be a clash.

I am a little worried about the *should*, because I had a look at the code and could see that an arriving packet in `ZigbeeNetworkManager#receiveCommand()` leads to the call `commandNotifier.notifyCommandListeners(command)`.

The `ZigBeeNetworkDiscoverer` is one of these `CommandListeners` and creates new nodes when packets arrive.

In this chain of commands I could not see a check on the network state, i.e. that the state is ONLINE before doing anything.

---

This makes me wonder if it is even feasible to implement such a check. A check of ONLINE is simple, but aren't there packets which should also be handled while the network manger is still in INITIALIZING state? Then we cannot implement such a check.

In my opinion, the safest way would be to have the new API method, which adds nodes only if they do not exist, because we then do not have to deal with all the different network states.

cdjackson:
I don’t think there should be ANY packets received when the network is in initialising state. In this state the network is typically offline (at the end of this state it will transition to online) so it should not be possible to receive packets. 

We can add a hard check for this if you prefer - it is already handled I think like this in the ember handler.

triller-telekom:
> I don’t think there should be ANY packets received when the network is in initialising state

If that is true, we do not have to change anything.

But while looking at the notification of the listeners I noticed an inconsistency that `addNode()` does NOT inform the listeners if the network is OFFLINE, but `removeNode()` and `updateNode()` do.

I have created a PR to fix this in 1.2.0:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/615

cdjackson:
Thanks. I think this was added only in addNode as this is the normal route - it's not normal for them to update or be deleted. You're right though that it's best to do this in all cases for consistency (thanks).

triller-telekom:
I have a safety guard added with PR https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/624

triller-telekom:
With https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/590 the networkmanager can be set into a reinitialization state via `reinitialize` and in that nodes can be added or removed.

So this PR can be closed.

Issue #604, opened 2019-04-25
Ember SPI transaction does not match Request and Response correctly


triller-telekom:
During initialization of an Ember chip connected via SPI we have observed the following:

In `ZigBeeDongleEzsp.initialize()`, we see

```
2019-04-25T09:33:30.116+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp] - Ember local IEEE Address is 000D6F000F24146C
```

But we do not reach the debug output ""EZSP dongle initialize done: Initialised {}" at the end of that method.

Digging deeper into the issue we saw that we send the `EzspGetCurrentSecurityStateRequest`:

```
2019-04-25T09:33:32.563+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler] - TX EZSP: EzspGetCurrentSecurityStateRequest []
2019-04-25T09:33:32.564+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler] - --> TX SPI frame: FE 03 4C 00 69 
2019-04-25T09:33:32.564+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp] - RX EZSP: EzspGetNetworkParametersResponse [status=EMBER_SUCCESS, nodeType=EMBER_COORDINATOR, parameters=EmberNetworkParameters [extendedPanId=752C921BEE6FA922, panId=28456, radioTxPower=0, radioChannel=11, joinMethod=EMBER_USE_MAC_ASSOCIATION, nwkManagerId=0, nwkUpdateId=0, channels=07FFF800]]
```

and shortly afterwards a `EzspGetNodeIdResponse` arrives:

```
2019-04-25T09:33:32.678+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler] - <-- RX SPI frame: FE 05 4B 80 27 00 00 
[...]

PR #622, opened 2019-05-05
Synchronise initialisation of SPI pollingTimer

cdjackson:
This addresses some potential synchronisation issues in the Ember SPI handler.

1) the stop and starting of the polling timer should probably be synchronised although I can't really see a mechanism that could leave the timer stopped.

2) ```lastSentFrame``` is partially synchronised by ```outputFrameSynchronisation```. There are a couple of instances where this changes or is checked outside of this synchronisation.

Both of these are potentially related to #604 since if the polling timer was stopped, or ```lastSentFrame``` was not set to null after a frame is received, then polling would stop, and this could cause the issue in #604.

@triller-telekom - what do you think? I'm not confident this is the cause of #604, but I think these should be synchronised, but I'd welcome your thoughts...


cdjackson:
Agreed - I'm also not sure it will fix the other issue ;)

PR #624, opened 2019-05-06
Networkmanager should ignore commands if the network is not ONLINE

triller-telekom:
This is a safety guard so incoming pakets do not interfere with
(re-)initialization of the ZigbeeNetworkManager.


triller-telekom:
@cdjackson Can you please have a look?

Also do you have a rough timeline when you want to release the 1.2 branch? Since we have merged the manufacturer specific clusters this morning, I wonder when there is a 1.2 release so we can provide the PR for the openHAB binding.

cdjackson:
Sorry - I did look at this earlier. I'm fine with this, but haven't yet merged it as I wanted to do it tonight - I think it will cause a conflict with #623 so I want to think about the best way to merge these ;)

cdjackson:
Oh, and regarding 1.2 merge, I will do this in the next day or two (ie this week).

triller-telekom:
Let me know if I should rebase anything from this PR if you run in trouble with #623. Or just feel free to pick my few lines via copy&paste into your branch. I am fine with that too.

PR #635, opened 2019-05-19
Change handling of SPI transaction timeout

cdjackson:
@triller-telekom can you take a look at this and see if it works on your SPI system. This increases the timeout to 5 seconds to give us a good chance of continuing, and removes the retry, replacing it with a call to set the system to offline.

#604 

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java
@@ -50,6 +50,10 @@ import com.zsmartsystems.zigbee.transport.ZigBeePort;
  * return with the completed result.
  * <p>
  * AN711: EZSP-SPI Host Interfacing Guide
+ * <p>
+ * This handler does not retry messages since all commands must receive a response in the SPI synchronous system.
+ * A timer is used to time-out any commands that don't receive a response and the link is taken offline. The upper
+ * layer must restart the connection.
  *
  * @author Chris Jackson
  *
@@ -59,12 +63,6 @@ public class SpiFrameHandler implements EzspProtocolHandler {
      * The logger.
      */
     private final Logger logger = LoggerFactory.getLogger(SpiFrameHandler.class);
-
-    /**
-     * Maximum number of consecutive timeouts allowed while waiting to receive the response
-     */
-    private final int ACK_TIMEOUTS = 4;
-    private int retries = 0;
 
     private final int SPI_FLAG_BYTE = 0xA7;
 
@@ -98,9 +96,9 @@ public class SpiFrameHandler implements EzspProtocolHandler {
     private Object outputFrameSynchronisation = new Object();
 
     /**
-     * Response timeout. The number of milliseconds to wait for a response before resending the request.
+     * Response timeout. The number of milliseconds to wait for a response before timing out.
      */
-    private int receiveTimeout = 500;
+    private int receiveTimeout = 5000;
 
     /**
      * Callback polling rate in milliseconds.
@@ -140,8 +138,7 @@ public class SpiFrameHandler implements EzspProtocolHandler {
     private Thread parserThread = null;
 
     /**
-     * Flag reflecting that parser has been closed and parser parserThread
-     * should exit.
+     * Flag reflecting that parser has been closed and parser parserThread should exit.
      */
     private boolean closeHandler = false;
 
@@ -297,7 +294,6 @@ public class SpiFrameHandler implements EzspProtocolHandler {
     private void spiTransactionComplete() {
         synchronized (outputFrameSynchronisation) {
             lastFrameSent = null;
-            retries = 0;
         }
         stopRetryTimer();
     }
@@ -588,7 +584,7 @@ public class SpiFrameHandler implements EzspProtocolHandler {
         stopRetryTimer();
 
         // Create the timer task
-        timerTask = new RetryTimer();
+        timerTask = new CancelTimer();
         timer.schedule(timerTask, receiveTimeout);
     }
 
@@ -600,35 +596,16 @@ public class SpiFrameHandler implements EzspProtocolHandler {
         }
     }
 
-    private class RetryTimer extends TimerTask {
+    private class CancelTimer extends TimerTask {
         @Override
         public void run() {
-            synchronized (outputFrameSynchronisation) {
-                logger.debug("SPI Timer task [{}].  {}", retries, frameToString(lastFrameSent));
+            logger.debug("SPI Timer task triggered.");
 
-                // Resend the last message sent if there is one
-                if (lastFrameSent == null) {
-                    return;
-                }
+            // We should alert the upper layer so they can reset the link
+            frameHandler.handleLinkStateChange(false);
 
-                if (retries++ > ACK_TIMEOUTS) {
-                    // Too many retries.
-                    // We should alert the upper layer so they can reset the link?
-                    frameHandler.handleLinkStateChange(false);
-
-                    logger.debug("Error: number of retries exceeded [{}].", retries);
-                    // drop message
-                    lastFrameSent = null;
-                    retries = 0;
-                    return;
-                }
-
-                try {
-                    outputFrame(lastFrameSent);
-                } catch (Exception e) {
-                    logger.warn("Caught exception while attempting to retry message in SpiRetryTimer", e);
-                }
-            }
+            // Close the handler to cancel outstanding transactions and prevent new transactions
+            close();
         }
     }
 
--- a/com.zsmartsystems.zigbee.dongle.ember/src/test/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandlerTest.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/test/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandlerTest.java
@@ -7,7 +7,10 @@
  */
 package com.zsmartsystems.zigbee.dongle.ember.internal.spi;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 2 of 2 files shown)


com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java
```diff
@@ -603,32 +599,10 @@ private synchronized void stopRetryTimer() {
...
-                }
-
-                try {
-                    outputFrame(lastFrameSent);
-                } catch (Exception e) {
-                    logger.warn("Caught exception while attempting to retry message in SpiRetryTimer", e);
-                }
-            }
+            // We should alert the upper layer so they can reset the link
+            frameHandler.handleLinkStateChange(false);
```
triller-telekom: I fear that the restart here will not work. When the problem from https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/604 occurs, we are still initializing.

This means that in

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/5825dcff62edc75e02ced49168d07c1b3078d447/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java#L720-L722

the variable `initialised` is still `false` and thus we will `return` instead of setting the transport to `OFFLINE`.

triller-telekom:
I just tried this commit on my device which uses SPI, but unfortunately the bridge does not come ONLINE either.

What I see is this exception:

```
2019-05-20T16:39:53.382+0200   446  WARNING        [symbolicName=org.openhab.binding.zigbee.ember, className=org.openhab.binding.zigbee.ember.handler.ZigBeeSerialPortSPI] - IO exception when writing bytes
java.io.IOException: Device or resource busy in writeArray
	at gnu.io.RXTXPort.writeArray(Native Method)
	at gnu.io.RXTXPort$SerialOutputStream.write(RXTXPort.java:1167)
	at org.openhab.binding.zigbee.ember.handler.ZigBeeSerialPortSPI.write(ZigBeeSerialPortSPI.java:145)
	at com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler.outputFrame(SpiFrameHandler.java:537)
	at com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler.outputFrame(SpiFrameHandler.java:517)
	at com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler.connect(SpiFrameHandler.java:558)
	at com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp.initialiseEzspProtocol(ZigBeeDongleEzsp.java:917)
	at com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp.initialize(ZigBeeDongleEzsp.java:318)
	at com.zsmartsystems.zigbee.ZigBeeNetworkManager.initialize(ZigBeeNetworkManager.java:334)
	at org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler.initialiseZigBee(ZigBeeCoordinatorHandler.java:423)
	at org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler.lambda$0(ZigBeeCoordinatorHandler.java:504)
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(Unknown Source)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)
```

whereas `org.openhab.binding.zigbee.ember.handler.ZigBeeSerialPortSPI` is a custom class from our solution and the implementation of `write()` looks like:

```
@Override
    public void write(int value) {
        byte[] outBuffer = getPacket(value);
        if (outBuffer == null || outputStream == null) {
            return;
        }
        try {
            outputStream.write(outBuffer);
        } catch (IOException e) {
            logger.warn("IO exception when writing bytes", e);
        }
    }
```

Interesting though that in `com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler.outputFrame(SpiFrameHandler.java:537)` `port.write(SPI_FLAG_BYTE);` is called which results in the exception. However, 2 lines before `port.write(outByte);` was called which apparently went smooth.

I am pretty sure that before this PR we had this exception as well, but the retry mechanism made it work afterwards.

I don't know whats wrong, i.e. why we get the `Device or resource busy` Exception, but a retry mechanism would be a good idea. There might even be flaws in the underlying serialport library, which we could overcome with a retry mechanism.

For now, I don't know whats wrong. @cdjackson maybe you have an idea what might happen there.

cdjackson:
I probably wont have time to look at this this week. My thinking though is that if there is a problem with the underlying serial implementation, then we probably should try and fix that in the OH handler rather than adding a retry here which is a more general implementation. I'm not super sure how we would handle retries - as we've seen, they are going to cause problems later on - maybe we get away with it in this case due to the command not being sent to the NCP, but in most cases, where we assume that the command is sent, then it will fail as we will get out of sync.

I will try and look at this if I can get some time, but it might be next week.

cdjackson:
@triller-telekom can you give this another try please.

What I've added is to close the SPI handler when the timeout occurs. What this ought to do is the kill off all outstanding transactions. So, in this case we have the following -:

```
        // Connect to the ASH handler and NCP
        frameHandler.start(serialPort);
        frameHandler.connect();

        EmberNcp ncp = getEmberNcp();

        // We MUST send the version command first.
        // Any failure to respond here indicates a failure of the ASH or EZSP layers to initialise
        EzspVersionResponse version = ncp.getVersion(4);
        if (version == null) {
            logger.debug("Version returned null. ASH/EZSP not initialised.");
            return false;
        }
```

The ```connect``` queues the first frame which will fail after 5 seconds, and this should cause the version response to return null as the transaction should now fail. This in turn returns ```false``` here, which in turn will return ```ZigBeeStatus.COMMUNICATION_ERROR``` from the ```ZigBeeDongleEzsp.initialize()``` method. Then there is the question of what happens in the binding - this will need to handle the failure from ```initialize``` and do a retry - I've not checked to see if this happens or not, but we can of course look at that...

I think this is a reasonable way to handle this - so long as it solves the issue that you see in your driver... If the exception happens every time the port is opened and first written to, then maybe closing and re-opening the port here will not help, but I think if that is the case, then it might be better to handle that in your SPI handler as I think that will be a specific issue with your SPI implementation.

WDYT?

triller-telekom:
Your current implementation is OK in general, and indeed the openHAB binding will trigger its reconnectJob. So this PR is good to be merged as it improves the situation.

However, we observed that between sending the initial get version request (i.e. first ever command) and the `SPI_FLAG_BYTE` we need a delay. We experimented with 300ms and it seemed fine. By briefly scanning some SPI specification documents we noticed that there are several "waiting times", up to 300ms, so it seems like a reasonable value.


SpiFrameHandler:
```
// Synchronize this method to ensure a packet gets sent as a block
    private void outputFrame(int[] outputData, boolean tracelogFrame) {
        synchronized (outputFrameSynchronisation) {
            lastFrameSent = outputData;

            String logMessage = String.format("--> TX SPI frame: %s", frameToString(outputData));
            if (tracelogFrame) {
                logger.trace(logMessage);
            } else {
                logger.debug(logMessage);
            }

            // Send the data
            for (int outByte : outputData) {
                logger.trace("SPI TX: {}", String.format("%02X", outByte));
                port.write(outByte);
            }

            if (Arrays.equals(requestSpiVersion, outputData)) {
                logger.debug("Sleeping HACK active");
                try {
                    Thread.sleep(300);
                    logger.debug("Sleeping HACK done");
                } catch (InterruptedException e) {
                    logger.debug("Exception while sleeping HACK: {}", e);
                    // TODO Auto-generated catch block
                    e.printStackTrace();
                }
            }

            port.write(SPI_FLAG_BYTE);

            startRetryTimer();
        }
    }
```

Before this change it was almost 100% reproducible with your PR that the network does not come online at all and the reconnectJob is trying to re-initialize indefinitely.

We are not sure what the correct way is to implement such a delay. In the code we check whether the to be send data are the version request and if so we delay the frame termination byte.

@cdjackson Since you have a deeper knowledge of the code, do you know a better place on how to implement this type of delay mechanism?

cdjackson:
Thanks for the feedback Stefan. I will take a look at this tonight (I’m East Coast US time at the moment).

If I understand correctly though, I don’t think we need the delay - it sounds to me (and please correct me if my understanding is wrong) that the binding is not waiting for the response to the getVersion request? Is that correct? I will check this myself later, but if you can comment on this, it might be as simple as ensuring we are treating the getVersion as a proper transaction - which should in any case be the way it should work.

triller-telekom:
> the binding is not waiting for the response to the getVersion request

That's correct.

BUT, we put the delay INTO a command (the get version request), so between the command byte and the SPI termination byte.

It looks like that the chip needs a few milliseconds to process the command byte (if it is a `SPI_CMD_SPIVERSION=0x0A`), before it can receive and process the `SPI_FLAG_BYTE=0xA7` byte.

So wrapping the current request into a transaction won't help, the only thing which does help is inserting that delay between the command and the termination byte.

Please have a look at the code example above to see where exactly the delay has to be inserted.

And as stated above, I tried this at least 10 times in a row and it worked consistently, whereas before there was a >50% chance that it did not work.

cdjackson:
Good point and that seems strange. I wonder if this is really an issue with the chip (ie the NCP has to wait) which I really think is unlikely since this is a single SPI packet - or - if this is actually a problem with the serial to SPI implementation in your gateway?

If this is a problem with the NCP, and the NCP really needs this delay, then I’m happy to put a fix in here (not sure what/where yet) - however if this is an issue with the driver, then I think we should fix this a different way (ie in your SPI handler). My guess is that this isn’t an NCP issue, but let’s see…

If this needs to be fixed in the SPI handler, one option that might work is to simply add this to the routine that opens the port - ie send the SPI_CMD_SPIVERSION, sleep(300), then SPI_FLAG_BYTE in the port open method. This might be a good option for you to try? WDYT? If we find this is really an NCP issue, then I will discuss with Silabs and if required we can certainly add the fix in the libraries.

I’m in the process of sorting out an SPI interface that I can use on my Mac so will be able to test this soon I hope.

triller-telekom:
Thanks for having a look at this problem.

I took your suggestions and tried the following: In our `ZigBeeSerialPortSPI` implementation, right after the port is opened in `open()`, I send the `SPI_CMD_SPIVERSION=0x0A`, wait 300ms and send `SPI_FLAG_BYTE=0xA7`. This works too.

So we still don't know whether its the chip or the serial implementation, but in case its the serial implementation we can add the workaround in our code.

Meanwhile feel free to merge this PR.

cdjackson:
Thanks Stefan.

I’ll merge this later, and I will check with Silabs if they have any comment on the 300mS delay and we can make a decision on where to put any fix.

PR #646, opened 2019-05-29
Added SHUTDOWN->OFFLINE to allowed transitions in ZigBeeNetworkManager

### LINKED ISSUE #644: NPE in ZigBeeNetworkManager on network shutdown
In `ZigBeeNetworkManager.setTransportState(ZigBeeTransportState)` a NPE is thrown when processing this statement
```
if (!validTransportStateTransitions.get(networkState).contains(state)) {
```
In that moment the value of `networkState` is `SHUTDOWN` which is not contained in the collection of valid state transitions, and thus `validTransportStateTransitions.get(networkState)` returns `null` followed by a NPE. My proposal is to add a transition `SHUTDOWN -> OFFLINE` to the collection. I will prepare a PR.

TomTravaglino:
Resolves #644 

Signed-off-by: Tommaso Travaglino <tommaso.travaglino@telekom.de>

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -264,7 +264,7 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
         transitions.put(ZigBeeNetworkState.INITIALISING, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
         transitions.put(ZigBeeNetworkState.ONLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
         transitions.put(ZigBeeNetworkState.OFFLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
-        transitions.put(ZigBeeNetworkState.SHUTDOWN, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
+        transitions.put(ZigBeeNetworkState.SHUTDOWN, new HashSet<>());
 
         validTransportStateTransitions = Collections.unmodifiableMap(new HashMap<>(transitions));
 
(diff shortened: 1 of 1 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -264,6 +264,7 @@ public ZigBeeNetworkManager(final ZigBeeTransportTransmit transport) {
         transitions.put(ZigBeeNetworkState.INITIALISING, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
         transitions.put(ZigBeeNetworkState.ONLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
         transitions.put(ZigBeeNetworkState.OFFLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
+        transitions.put(ZigBeeNetworkState.SHUTDOWN, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
```
cdjackson: I'm not sure that this is correct - this will allow the system to be restarted, and that is not allowed. How about adding an empty set here instead?

TomTravaglino: This means there is no allowed transition from SHUTDOWN to any state? In this case I would suggest to not add any transition to the collection and instead change the method `setTransportState(final ZigBeeTransportState state)` to do a null check. WDYT?

cdjackson: > This means there is no allowed transition from SHUTDOWN to any state?

That's correct - as per the state transition diagram -:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/master/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/doc-files/ZigBeeNetworkStateTransitions.png

image


> In this case I would suggest to not add any transition to the collection and instead change the method setTransportState(final ZigBeeTransportState state) to do a null check. WDYT?

I don't mind too much. Just adding the empty Set seems ok to me - it's (arguably) cleaner, and does make it clear if someone looks at all the allowed state transitions, that SHUTDOWN has no allowable states.

If you think the null check is simpler though, I don't mind, but I think the code will look more cluttered ;)

TomTravaglino: OK. Then I'll just add an empty set.

TomTravaglino: I updated the PR

cdjackson:
Thanks @TomTravaglino

PR #643, opened 2019-05-28
Telegesis: Delay the start of the network state polling by 15 seconds

triller-telekom:
Fixes #640

Signed-off-by: Stefan Triller <stefan.triller@telekom.de>

PR #647, opened 2019-05-29
Refactor Telegesis state polling

cdjackson:
Moves the Telegesis state polling from the frame handler to the higher level code. Polling is now only performed if the dongle is online.
This is consistent with the implementation in the Ember dongle handler.


cdjackson:
@triller-telekom did you have a chance to look at this to see if it solves your issue?

triller-telekom:
@cdjackson Sorry that I get back to you this late. I just tested it on a slow machine where this issue occurred quite often. With this PR it didn't happen so far.

Un-plugging and re-plugging the stick in openHAB is still working as well, so the bridge goes offline and back online, so the retry logic wasn't broken with this PR.

Also code-wise it looks alright to me, since its now aligned with the ember implementation.

Thanks for implementing this. I will close #643.

cdjackson:
Thanks @triller-telekom

Issue #651, opened 2019-06-07
NodeServiceDiscovery is not removed when node is removed


triller-telekom:
There seems to be an issue with a mapping of nodes to IEEE addresses in the discovery somewhere. We see that our discovery never completes, i.e. there is no DiscoveryResult created for openHAB.

Somehow we have ONE IEEE=00124B000420A2B5 address, but TWO nodes: 42581 and 33788. We see in the sniffer that 
42581 is the correct node which is responding to our requests.

Before we started the scan for new devices:

```
$zigbee nodes

Network  Addr  IEEE Address       Logical Type   EP   Profile                    Device Type      Manufacturer     Model
      0  0000  000D6F000AF902EA   COORDINATOR

```

After the scan with no DiscoveryResult for openHAB:

```
$zigbee nodes

Network  Addr  IEEE Address       Logical Type   EP   Profile                    Device Type      Manufacturer     Model
      0  0000  000D6F000AF902EA   COORDINATOR
  33788  83FC  00124B000420A2B5   UNKNOWN

```

So there is a wrong node in the NetworkManager somehow.

The problem seems to be within the NodeServiceDiscoveryTask:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/36f8412da5f6e2f812cd42a7f0b02768772de427/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java#L585-L619

There is the call to `networkManager.updateNode(node)` which overrides the content of the current node with the information from the node given as an argument. Whenever one of the discovery subtasks finishes, it updates the node. However, for the "wrong" node there are timeouts and thus no values available.

When there are timeouts for the non existing 33788 node, we will overwrite the already successfully discovered properties for 42581, because they both have the same IEEE address (That is because `networkManager.updateNode(node)` does: `currentNode = networkNodes.get(node.getIeeeAddress());`.

@cdjackson: Would be nice if you could have a look please, since you are more familiar with the code.

The log was done using a telegesis Dongle on one of our very slow devices, so maybe timing could also play a role as a side effect here.

[logfile.txt](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/files/3266356/log3.txt)

cdjackson:
Please can you explain what you mean by -:

[...]

PR #652, opened 2019-06-11
Close the Node Service Discovery if the thread is running

cdjackson:
@triller-telekom this relates to issue #651. There is code in place to stop the discovery when a node is removed, however I see a corner case in that if a task is running when the node is removed, then it will be rescheduled.  Here I've added a flag to try and avoid this.

Maybe there's a better approach here so please feel free to comment.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
@@ -164,6 +164,8 @@ public class ZigBeeNodeServiceDiscoverer {
      */
     private final Queue<NodeDiscoveryTask> discoveryTasks = new LinkedList<NodeDiscoveryTask>();
 
+    private boolean closed = false;
+
     /**
      * Creates the discoverer
      *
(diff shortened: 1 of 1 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
```diff
@@ -618,6 +624,11 @@ public void run() {
                         break;
                 }
 
+                if (closed) {
```
triller-telekom: I wonder if we should do this check at the very beginning of the `run` method to avoid the race condition where we start the discovery and at the same time remove the node. Then we won't call the `requestXYZ()` methods.

cdjackson: The reason I put it after various methods is that is where the time is spent - ie this loop itself is very short from a time perspective (probably microseconds). However, the requests will take time (measured in tens of milliseconds, or even tens of seconds if the request times out). So, if we moved this check to the beginning, then it's fine if it gets caught, however, if it misses, and we start the requests, then we will end up rescheduling this.

Also, the shutdown method will stop the schedule, so in theory this should not run unless the run method was already executing when the shutdown was called. So, adding the check at the beginning of the method is probably not going to help at all.

triller-telekom: Good point, I agree that most time would be spend in the `requestXYZ()` calls.

Then the PR looks good to me.

cdjackson:
Thanks.

PR #653, opened 2019-06-11
Defaults the APS counter to -1 to indicate that it is not being used

cdjackson:
@triller-telekom FYI this should resolve the issue we saw in your logs relating to Telegesis messages being dropped due to duplicate APS counters.


cdjackson:
Thanks @triller-telekom

Issue #663, opened 2019-06-16
ASH: ERROR received (code 81). Disconnecting.


openhab-5iver:
Moving over from https://github.com/openhab/org.openhab.binding.zigbee/issues/440. There are lots of details in that issue.

The issue started when using 1.1.11 libraries, but also occurred using 1.1.6. As reported [in this comment](https://github.com/openhab/org.openhab.binding.zigbee/issues/440#issuecomment-502476354), the error has stopped occurring when not using a manual install of the binding and a patch to openhab-core. The high load issue in the current OH snapshots could just be revealing an underlying issue in the libraries.

cdjackson:
I'm still discussing this with Silabs support - it's slow going...

At the moment, when an error is received the ASH handler disconnects. What I will look at is to retry the RESET if we're not connected - that might help and we can try it in the absence of info from Silabs.

cdjackson:
I spent some time trying to replicate this. By frigging around with the ASH protocol and putting a hack in to stop sending frames after some time, I can make the dongle go into error mode -:

```
18:26:54.125  DEBUG  ASH: TX EZSP queue size: 1
18:26:54.125  DEBUG  ******* LETS BLOW THIS BABY SKY HIGH
18:26:54.452  DEBUG  <-- RX ASH frame: AshFrameData [frmNum=6, ackNum=1, reTx=true, data=79 80 FF 00 0B 02 FF FF]
18:26:54.453  DEBUG  ******* LETS BLOW THIS BABY SKY HIGH
18:26:57.579  DEBUG  <-- RX ASH frame: AshFrameError [version=2, errorCode=81, Error: Exceeded maximum ACK timeout count]
18:26:57.579  WARN   ASH: ERROR received (code 81).
18:26:57.579  DEBUG  ASH: ERROR received (code 81). Disconnecting.
```

Unfortunately though when I restart, the reset seems to work fine -:

```
18:27:47.566  DEBUG  --> TX ASH frame: AshFrameRst []
18:27:47.617  DEBUG  TX EZSP: EzspVersionRequest [desiredProtocolVersion=4]
18:27:47.624  DEBUG  ASH: TX EZSP queue size: 1
18:27:48.967  DEBUG  <-- RX ASH frame: AshFrameRstAck [version=2, resetCode=11, Reset: Software]
18:27:48.967  DEBUG  ASH: Connected
```

[...]

PR #666, opened 2019-06-21
Retry ASH reset if error received before connected

cdjackson:
With this PR the ASH layer should retry the ASH RESET command if the layer is not connected. This may help to to resolve the issue seen in #663 where the dongle is not responding to the reset if it's in the error state.


cdjackson:
Hmmm - PR 666 - not sure if I should merge this ;)

PR #670, opened 2019-06-29
Updated handling of adding nodes and associated notifications

### LINKED ISSUE #184: Mesh update node changed check not working
When the mesh update completes, it copies the new data into the node. It then does a check to see if the node has changed, and if so, it then sends the notification to listeners. Since the node is updated already, the check always fails and notifications are not sent.

The update of the node in the mesh update code should be looked at to improve this.

#183

### LINKED ISSUE #649: NetworkManager updateNode notifies listeners in a wrong way
I am currently debugging why openHAB as a client of this library starts a discovery for a node twice.

I dug down the code and stumbled across the `updateNode()` method of the `ZigbeeNetworkManager`:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/36f8412da5f6e2f812cd42a7f0b02768772de427/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java#L1369-L1421

The decision of whether to inform the listeners about an _added_ or _updated_  node is taken in line 1398. However this decision is based on the fact that the discovery for that particular node has been finished or not.

So whenever a node gets changed and its discovery has not been completed yet (this is a potentially long running thread in the background), this will result in a call to `listener.nodeAdded()` instead of `listener.nodeUpdated()`.

IMHO the decision about updated vs added should only be because of changed properties in the node itself.

In the case of openHAB, there a multiple discoveries started because it acts as a `nodeAdded()` listener and whenever a new node occurs, it starts to read out its zigbee features. if the above decision would be corrected and `nodeUpdated()` would be called instead of `nodeAdded()`, this issue would be resolved.

cdjackson:
This is a change to the way that nodes are added to the network manager, and the associated notifications.

Nodes are now added to the network through the ```updateNode``` method. This will send a ```nodeAdded``` on initial addition, or a ```nodeUpdated``` notification for subsequent updates.

The ```nodeAdded``` will not be sent immediately if the network is not ```ONLINE``` - it will be deferred until the network comes up.

This also fixes the sending of the ```nodeUpdated``` notification so that it is only sent if the node data has really changed. As part of this, we ensure that there are no changes to existing clusters, so that currently installed listeners are not lost, meaning that applications fail to receive notifications.

Closes #184
Closes #649 

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -238,7 +238,12 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
     /**
      * Map of allowable transport state transitions
      */
-    private final Map<ZigBeeNetworkState, Set<ZigBeeTransportState>> validTransportStateTransitions;
+    private final Map<ZigBeeTransportState, Set<ZigBeeTransportState>> validTransportStateTransitions;
+
+    /**
+     * Current state of the transport layer
+     */
+    private ZigBeeTransportState transportState = ZigBeeTransportState.UNINITIALISED;
 
     /**
      * Our local {@link IeeeAddress}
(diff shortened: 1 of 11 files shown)


cdjackson:
This pull request **introduces 1 alert** when merging 000092ad42530498bd7989d3ffa9a800116c81fa into baa37b23229b7dfc0e33df78b900897ed760685b - [view on LGTM.com](https://lgtm.com/projects/g/zsmartsystems/com.zsmartsystems.zigbee/rev/pr-40173c4c57ec6a24dae0612b83c44358c26f1c37)

**new alerts:**

* 1 for Reference equality test of boxed types

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

PR #671, opened 2019-06-30
Updated handling of adding nodes and associated notifications

### LINKED ISSUE #184: Mesh update node changed check not working

### LINKED ISSUE #649: NetworkManager updateNode notifies listeners in a wrong way

cdjackson:
This is a change to the way that nodes are added to the network manager, and the associated notifications are sent to consumers. This should make the notifications clearer and cleaner, but please note that this may require changes in the ```nodeAdded``` or ```nodeUpdated``` listeners in application code.

Nodes are now added to the network through the ```updateNode``` method. This will send a ```nodeAdded``` on initial addition, or a ```nodeUpdated``` notification for subsequent updates.

The ```nodeAdded``` will not be sent immediately if the network is not ```ONLINE``` - it will be deferred until the network comes up.

This also fixes the sending of the ```nodeUpdated``` notification so that it is only sent if the node data has really changed. As part of this, we ensure that there are no changes to existing clusters, so that currently installed listeners are not lost, meaning that applications fail to receive notifications.

This change required numerous changes throughout the code to better allow this change detection and consolidation of the ```ZigBeeNode``` fields.

There has also been some small changes to the startup state detection when the ```ZigBeeTransportState``` is updated - again this is all related to the node initialisation and update notifications.

Closes #184
Closes #649 

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -373,9 +373,7 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
             ZigBeeNode node = getNode(ieeeAddress);
             if (node == null) {
                 logger.debug("{}: Adding local node to network, NWK={}", ieeeAddress, nwkAddress);
-                node = new ZigBeeNode(this, ieeeAddress);
-                node.setNetworkAddress(nwkAddress);
-
+                node = new ZigBeeNode(this, ieeeAddress, nwkAddress);
                 updateNode(node);
             }
         }
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -238,7 +238,12 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
     /**
      * Map of allowable transport state transitions
      */
-    private final Map<ZigBeeNetworkState, Set<ZigBeeTransportState>> validTransportStateTransitions;
+    private final Map<ZigBeeTransportState, Set<ZigBeeTransportState>> validTransportStateTransitions;
+
+    /**
+     * Current state of the transport layer
+     */
+    private ZigBeeTransportState transportState = ZigBeeTransportState.UNINITIALISED;
 
     /**
      * Our local {@link IeeeAddress}
@@ -258,13 +263,12 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
     public ZigBeeNetworkManager(final ZigBeeTransportTransmit transport) {
         databaseManager = new ZigBeeNetworkDatabaseManager(this);
 
-        Map<ZigBeeNetworkState, Set<ZigBeeTransportState>> transitions = new ConcurrentHashMap<>();
-        transitions.put(ZigBeeNetworkState.UNINITIALISED,
+        Map<ZigBeeTransportState, Set<ZigBeeTransportState>> transitions = new ConcurrentHashMap<>();
+        transitions.put(ZigBeeTransportState.UNINITIALISED,
                 new HashSet<>(Arrays.asList(ZigBeeTransportState.INITIALISING, ZigBeeTransportState.OFFLINE)));
-        transitions.put(ZigBeeNetworkState.INITIALISING, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
-        transitions.put(ZigBeeNetworkState.ONLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
-        transitions.put(ZigBeeNetworkState.OFFLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
-        transitions.put(ZigBeeNetworkState.SHUTDOWN, new HashSet<>());
+        transitions.put(ZigBeeTransportState.INITIALISING, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
+        transitions.put(ZigBeeTransportState.ONLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
+        transitions.put(ZigBeeTransportState.OFFLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
 
         validTransportStateTransitions = Collections.unmodifiableMap(new HashMap<>(transitions));
 
@@ -372,7 +376,7 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
                 node = new ZigBeeNode(this, ieeeAddress);
                 node.setNetworkAddress(nwkAddress);
 
-                addNode(node);
+                updateNode(node);
             }
         }
     }
@@ -596,6 +600,7 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
      */
     public void executeTask(Runnable runnableTask) {
         if (networkState != ZigBeeNetworkState.ONLINE) {
+            logger.debug("ZigBeeNetworkManager executeTask: not executing task while {}", networkState);
             return;
         }
         executorService.execute(runnableTask);
@@ -610,6 +615,7 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
      */
     public ScheduledFuture<?> scheduleTask(Runnable runnableTask, long delay) {
         if (networkState != ZigBeeNetworkState.ONLINE) {
+            logger.debug("ZigBeeNetworkManager scheduleTask: not scheduling task while {}", networkState);
             return null;
         }
         return executorService.schedule(runnableTask, delay, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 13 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
```diff
@@ -459,21 +459,9 @@ private boolean getAssociatedNodes(final int networkAddress) throws InterruptedE
...
-            node.setNodeState(ZigBeeNodeState.ONLINE);
-            node.setNetworkAddress(networkAddress);
-            networkManager.updateNode(node);
-            return;
-        }
-
-        node = new ZigBeeNode(networkManager, ieeeAddress);
+        ZigBeeNode node = new ZigBeeNode(networkManager, ieeeAddress);
+        node.setNodeState(ZigBeeNodeState.ONLINE);
         node.setNetworkAddress(networkAddress);
```
triller-telekom: I wonder, while you are changing the API anyway, if its worth to make this `networkAddress` a required parameter for the constructor. Some places in the code already rely on it and fail with a NPW if the `networkAddress` is not set on the node. WDYT?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
```diff
@@ -195,24 +201,30 @@ private void startDiscovery(Set<NodeDiscoveryTask> newTasks) {
...
             if (newTasks.isEmpty()) {
-                logger.debug("{}: Node SVC Discovery: has no tasks to perform", node.getIeeeAddress());
+                logger.debug("{}: Node SVC Discovery: has no new tasks to perform", node.getIeeeAddress());
                 return;
             }
 
-            boolean startWorker = discoveryTasks.isEmpty();
+            // Check the current list of tasks to decide if we need to start the worker
+            // This prevents restarting if we didn't add new tasks, which might overload the system
+            boolean startWorker = discoveryTasks.isEmpty() | (futureTask == null);
```
triller-telekom: I think booleans should not be `or`ed with the bit-wise operator `|` but rather the `||`, so the compiler can complain once one side changes to a non-boolean by accident. Otherwise the compiler will really use the bit-wise or which is not intended here I guess.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -258,13 +263,12 @@
...
+        Map<ZigBeeTransportState, Set<ZigBeeTransportState>> transitions = new ConcurrentHashMap<>();
+        transitions.put(ZigBeeTransportState.UNINITIALISED,
                 new HashSet<>(Arrays.asList(ZigBeeTransportState.INITIALISING, ZigBeeTransportState.OFFLINE)));
-        transitions.put(ZigBeeNetworkState.INITIALISING, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
-        transitions.put(ZigBeeNetworkState.ONLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
-        transitions.put(ZigBeeNetworkState.OFFLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
-        transitions.put(ZigBeeNetworkState.SHUTDOWN, new HashSet<>());
+        transitions.put(ZigBeeTransportState.INITIALISING, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
+        transitions.put(ZigBeeTransportState.ONLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.OFFLINE)));
+        transitions.put(ZigBeeTransportState.OFFLINE, new HashSet<>(Arrays.asList(ZigBeeTransportState.ONLINE)));
```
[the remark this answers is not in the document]
cdjackson: Yes - you miss the point here - this is ZigBeeTransportState - not ZigBeeNetwork state. ZigBeeNetworkState.SHUTDOWN doesn't exist.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -1011,15 +1018,19 @@ public void removeNetworkStateListener(ZigBeeNetworkStateListener stateListener)
     public synchronized void setTransportState(final ZigBeeTransportState state) {
         logger.debug("ZigBeeNetworkManager transport state updated to {}", state);
 
-        if (!validTransportStateTransitions.get(networkState).contains(state)) {
-            logger.debug(
-                    "Ignoring invalid transport state transition in ZigBeeNetworkState.{} by ZigBeeTransportState.{}",
-                    networkState, state);
-            return;
+        // Filter out unwanted transport state changes
+        synchronized (validTransportStateTransitions) {
```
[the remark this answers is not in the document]
cdjackson: Since ```validTransportStateTransitions``` is final it can be used for synchronisation. It is not the map itself I'm worried about synchronising - it's the change to ```transportState``` - I can't synchronised on ```transportState``` as it is not a final field.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
cdjackson: I guess we could, but we don't always know the NWK address when the node is instantiated. What is REQUIRED is the EUI. I've not checked, but I suspect if we force the NWK to be provided in the constructor that it will make other areas more complex where NWK is not known.

I'll have a look at this though - I guess we could add an extra constructor at least.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
cdjackson: Thanks.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
cdjackson: I've added a new constructor that also takes the NWK address, and have changed applicable code to use this.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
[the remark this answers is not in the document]
triller-telekom: Ah OK, thanks for the explanation, that certainly makes sense.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
triller-telekom: Thanks

cdjackson:
Thanks for the review @triller-telekom. Have you run this in your system and confirmed it's ok? I've given it quite a lot of testing over the weekend, and another user has also confirmed it is working and resolves some issues they have so I think this is good to go - I just wanted to be sure that you'd also actually run it (I assume so, but wanted to confirm).

triller-telekom:
I just had the chance to actually test it, before I was just looking at the code sorry.

Unfortunately I have to say that I now run in some trouble in the openHAB binding. If I run the binding with the commits from https://github.com/openhab/org.openhab.binding.zigbee/pull/454/commits I only get a generic discovery result which shows the mac address of the device. I do not see the name and thus there can also be no static thing-type matching.

However, I can pair the generic device and it works. But this change certainly breaks the static thing-type matching.

With this PR here there has to be an additional commit inhttps://github.com/openhab/org.openhab.binding.zigbee/pull/454/commits as well to react on the `nodeUpdated` so the name and vendor of the device can be shown as a result in the inbox.

cdjackson:
Thanks Stefan. This isn’t exactly unexpected as we discussed in https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/649#issuecomment-498980996 <https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/649#issuecomment-498980996>. Once we can get the openHAB binding working in the IDE I’ll give this a test and update to resolve this.

PR #675, opened 2019-07-05
Ignore unknown (manufacturer specific) clusters instead of crashing with NPE

### LINKED ISSUE #674: Manufacturer specific clusters lead to NPE during initialization
Had had a Centralite 3310-D temperature and humidity device paired in our version of the openHAB binding, which uses manufacturer specific clusters.

If the binding serializes the data onto the disk and then openHAB is restarted, it reads out that serialized data again, however this will result in an NPE:

```
java.lang.NullPointerException
at com.zsmartsystems.zigbee.ZigBeeEndpoint.setDao(ZigBeeEndpoint.java:504)
at com.zsmartsystems.zigbee.ZigBeeNode.setDao(ZigBeeNode.java:783)
at com.zsmartsystems.zigbee.database.ZigBeeNetworkDatabaseManager.startup(ZigBeeNetworkDatabaseManager.java:135)
at com.zsmartsystems.zigbee.ZigBeeNetworkManager.initialize(ZigBeeNetworkManager.java:334)
at org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler.initialiseZigBee(ZigBeeCoordinatorHandler.java:427)
at org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler.lambda$0(ZigBeeCoordinatorHandler.java:508)
at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
at java.util.concurrent.FutureTask.run(Unknown Source)
at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(Unknown Source)
at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
at java.lang.Thread.run(Unknown Source)
```

The problem is here: https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/58ae488c4612a4977227b07920466838647a6542/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeEndpoint.java#L297-L303

ZSS does not have any information about that cluster and thus it cannot restore information about it and thus returns `null` here, which is correct. However in our case the openHAB binding (or for another solution on top of ZSS) does know the manufacturer specific clusters and will add the information to the ZSS node/endpoint upon initialization by themselves.

My proposal to address this would be to simply skip adding the cluster, if `getClusterClass(int clusterId)` returns `null`. I will provide a PR for that.

triller-telekom:
Fixes #674

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeEndpoint.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeEndpoint.java
@@ -501,15 +501,23 @@ public class ZigBeeEndpoint {
         if (dao.getInputClusterIds() != null) {
             for (ZclClusterDao clusterDao : dao.getInputClusters()) {
                 ZclCluster cluster = getClusterClass(clusterDao.getClusterId());
-                cluster.setDao(clusterDao);
-                inputClusters.put(clusterDao.getClusterId(), cluster);
+                if (cluster != null) {
+                    cluster.setDao(clusterDao);
+                    inputClusters.put(clusterDao.getClusterId(), cluster);
+                } else {
+                    logger.debug("Unknown input cluster found with id={}, will skip it", clusterDao.getClusterId());
+                }
             }
         }
         if (dao.getOutputClusterIds() != null) {
             for (ZclClusterDao clusterDao : dao.getOutputClusters()) {
                 ZclCluster cluster = getClusterClass(clusterDao.getClusterId());
-                cluster.setDao(clusterDao);
-                outputClusters.put(clusterDao.getClusterId(), cluster);
+                if (cluster != null) {
+                    cluster.setDao(clusterDao);
+                    outputClusters.put(clusterDao.getClusterId(), cluster);
+                } else {
+                    logger.debug("Unknown output cluster found with id={}, will skip it", clusterDao.getClusterId());
+                }
             }
         }
     }


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeEndpoint.java
```diff
@@ -501,15 +501,25 @@ public void setDao(ZigBeeEndpointDao dao) {
         if (dao.getInputClusterIds() != null) {
             for (ZclClusterDao clusterDao : dao.getInputClusters()) {
                 ZclCluster cluster = getClusterClass(clusterDao.getClusterId());
-                cluster.setDao(clusterDao);
-                inputClusters.put(clusterDao.getClusterId(), cluster);
+                if (cluster != null) {
+                    cluster.setDao(clusterDao);
+                    inputClusters.put(clusterDao.getClusterId(), cluster);
+                } else {
+                    logger.debug("Manufacturer specific input cluster found with id={}, will skip it",
```
cdjackson: Maybe we should just log "Unknown input cluster  found with id={}....." as we don't necessarily know it's a manufacturer specific cluster - just that it's a cluster that can't be created.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeEndpoint.java
```diff
@@ -501,15 +501,25 @@ public void setDao(ZigBeeEndpointDao dao) {
...
         if (dao.getOutputClusterIds() != null) {
             for (ZclClusterDao clusterDao : dao.getOutputClusters()) {
                 ZclCluster cluster = getClusterClass(clusterDao.getClusterId());
-                cluster.setDao(clusterDao);
-                outputClusters.put(clusterDao.getClusterId(), cluster);
+                if (cluster != null) {
+                    cluster.setDao(clusterDao);
+                    outputClusters.put(clusterDao.getClusterId(), cluster);
+                } else {
+                    logger.debug("Manufacturer specific output cluster found with id={}, will skip it",
```
cdjackson: Maybe we should just log "Unknown output cluster  found with id={}....." as we don't necessarily know it's a manufacturer specific cluster - just that it's a cluster that can't be created.

triller-telekom:
Thanks for the review. I agree that in those cases it might also be an invalid or really unknown cluster. Thus I have changed the logging. Please have a look again.

cdjackson:
Thanks Stefan.

PR #683, opened 2019-07-11
Update areAttributesManufacturerSpecific manufacturer specific check

cdjackson:
@triller-telekom you might like to check this since you are using the manufacturer specific attributes.


cdjackson:
Thanks Stefan.

The reason for the change is compatibility with Android. We can use Java8 compile time features, but not runtime features as only new quite versions of Android support these (unfortunately).

PR #687, opened 2019-07-17
Updates to SPI Driver

cdjackson:
This tidies up the SPI Handler and adds a few features -:
* Ensures that the timer is stopped when the handler is closed.
* Adds support for frames such as RESET which are sent by the SPI NCP code during initialisation. This required some minor refactoring of the receive handler to allow these frames to be supported.

@triller-telekom it would be great if you could give this a test on your system.


cdjackson:
Thanks @triller-telekom

PR #698, opened 2019-07-31
Fix ArrayIndexOutOfBoundsException in EzspFrame

### LINKED ISSUE #693: java.lang.ArrayIndexOutOfBoundsException in EzspFrame reintroduced see #586 and #641
PR #688 reintroduces the issue #586 which was fixed in #594, see

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/688/files#diff-29a5822c0fa62c8f8e2df9c53b307d9aL396

the try-catch block was removed, leading to the exception and a non working stack if there is an invalid frame.

This has happened for the second time now, see issue #641 and its fix #642.

triller-telekom:
Fix the issue in the autocoder too.

Fixes #693


triller-telekom:
Obsolete, I didn't see #697

PR #704, opened 2019-08-10
Use custom ThreadFactory for all Executors to provide named threads

cdjackson:
This creates a custom ```ZigBeeExecutors``` that uses a custom ```ThreadFactory``` so that all threads have more useful names.  All threads created explicitly have also been named.

This makes profiling and thread dumps more useful as all threads will now have a name that at least allows its source to be identified.

@triller-telekom any thoughts on this?

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNode.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNode.java
@@ -420,7 +420,7 @@ public class ZigBeeNode implements ZigBeeCommandListener {
         });
 
         // start the thread to execute it
-        new Thread(future).start();
+        new Thread(future, "UpdateBindingTable" + getIeeeAddress()).start();
         return future;
     }
 
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
@@ -787,7 +787,7 @@ public abstract class ZclCluster {
         });
 
         // Create the thread to execute it
-        new Thread(future, "DiscoverAttributes" + zigbeeEndpoint.getIeeeAddress()).start();
+        new Thread(future, "DiscoverAttributes-" + zigbeeEndpoint.getIeeeAddress()).start();
         return future;
     }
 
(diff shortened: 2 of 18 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNode.java
```diff
@@ -420,7 +420,7 @@ public ZigBeeStatus call() throws Exception {
         });
 
         // start the thread to execute it
-        new Thread(future).start();
+        new Thread(future, "UpdateBindingTable" + getIeeeAddress()).start();
```
triller-telekom: Isn't here a separator missing between the "UpdateBindingTable" and the address?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -787,7 +787,7 @@ public Boolean call() throws Exception {
         });
 
         // Create the thread to execute it
-        new Thread(future).start();
+        new Thread(future, "DiscoverAttributes" + zigbeeEndpoint.getIeeeAddress()).start();
```
triller-telekom: see above

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -897,7 +897,7 @@ public Boolean call() throws Exception {
         });
 
         // start the thread to execute it
-        new Thread(future).start();
+        new Thread(future, "DiscoverCommandsReceived" + zigbeeEndpoint.getIeeeAddress()).start();
```
triller-telekom: see above

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -1007,7 +1007,7 @@ public Boolean call() throws Exception {
         });
 
         // start the thread to execute it
-        new Thread(future).start();
+        new Thread(future, "DiscoverCommandsGenerated" + zigbeeEndpoint.getIeeeAddress()).start();
```
triller-telekom: see above

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNode.java
cdjackson: Actually, this was intended - I figured it was obvious enough without the separator, but you're right it is neater so I will update.

Thanks for the review.

PR #714, opened 2019-08-13
Allow late discovery of clusters on nodes for IAS

charles-toller:
The node discovery extension can take longer than expected to get the node descriptor, leading to the IAS extension missing the cluster when the node is added and ignoring it when the cluster is finally discovered. This adds a nodeUpdated hook to the extension to allow these nodes to be discovered and registered in the application.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/iasclient/ZigBeeIasCieExtension.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/iasclient/ZigBeeIasCieExtension.java
@@ -50,10 +50,23 @@ public class ZigBeeIasCieExtension implements ZigBeeNetworkExtension, ZigBeeNetw
 
     @Override
     public void nodeAdded(ZigBeeNode node) {
-        for (ZigBeeEndpoint endpoint : node.getEndpoints()) {
-            if (endpoint.getInputCluster(ZclIasZoneCluster.CLUSTER_ID) != null) {
-                endpoint.addApplication(new ZclIasZoneClient(networkManager, networkManager.getLocalIeeeAddress(), 0));
-                break;
+        if (node.isDiscovered()) {
+            for (ZigBeeEndpoint endpoint : node.getEndpoints()) {
+                if (endpoint.getInputCluster(ZclIasZoneCluster.CLUSTER_ID) != null && endpoint.getApplication(ZclIasZoneCluster.CLUSTER_ID) == null) {
+                    endpoint.addApplication(new ZclIasZoneClient(networkManager, networkManager.getLocalIeeeAddress(), 0));
+                    break;
+                }
+            }
+        }
+    }
+    @Override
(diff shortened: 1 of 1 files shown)


cdjackson:
I would like to better understand the issue as it should not be required to add a ```nodeUpdated``` handler here.

```nodeAdded``` should be called whenever new endpoints are added to the node - ```nodeUpdated``` should only be called when information such as neighbour tables etc change.

**Comment by charles-toller** [comment]
The discovery service obtains the node descriptor for the device, and upon finishing discovery which includes said node descriptor and endpoints, calls `nodeUpdated` [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/master/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java#L611), which is why I assumed the endpoints wouldn't come through nodeAdded. If this is unexpected behavior we can close this and fix the discovery extension instead.

**Comment by cdjackson** [comment]
That call that you reference to ```ZigBeeNetworkManager.updateNode``` is not the same as the ```ZigBeeNetworkNodeListener.nodeUpdated``` if that's what you are thinking.

```ZigBeeNetworkManager.updateNode``` checks to see if the node is updated, and either calls the ```ZigBeeNetworkNodeListener.nodeUpdated``` or ```ZigBeeNetworkNodeListener.nodeAdded``` notifications.

If there is an issue with the notifications, it is likely to be in the code that detects changes in the node and not in the discovery extension.

File: com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/iasclient/ZigBeeIasCieExtension.java [hunk 1]
```diff
@@ -50,10 +50,23 @@ public void extensionShutdown() {
 
     @Override
     public void nodeAdded(ZigBeeNode node) {
-        for (ZigBeeEndpoint endpoint : node.getEndpoints()) {
-            if (endpoint.getInputCluster(ZclIasZoneCluster.CLUSTER_ID) != null) {
-                endpoint.addApplication(new ZclIasZoneClient(networkManager, networkManager.getLocalIeeeAddress(), 0));
-                break;
+        if (node.isDiscovered()) {
```
triller-telekom: How about extracting this code into a separate method to avoid code duplication with the stuff you need to do in `nodeUpdated`?

cdjackson:
I am tempted to close this PR as it is not the way to address the problem. I would like to understand what the problem is, ensure that it is happening with the latest code, and then address it in the appropriate place which is not in the IAS handler, but in the detection of change and notification of events.

cdjackson:
@charles-toller please can you advise what you intend to do here?

charles-toller:
Chris, I think you're correct in closing this. I'll do more research on my
end, and open an issue with more info about the exact issue I'm seeing.

cdjackson:
Thanks @charles-toller . If you spot something that doesn't look right, then I'm happy to discuss it and we can find the right solution.

Issue #710, opened 2019-08-11
Enhance discovery of manufacturer specific clusters


cdjackson:
It could be possible to improve the current implementation of the manufacturer specific clusters. Currently, these are simply ignored when creating the ```ZigBeeEndpoint``` as they are unknown.

I propose to create a new class ```ZclCustomCluster``` which can be used as a placeholder for any clusters found in an endpoint that are not known by the framework. 

The implementation could (optionally?) use the ```discoverAttributes``` to add all supported attributes to the cluster to support reading and writing of attributes, configuration of binding/reporting etc without having to write a custom cluster class.

PR #718, opened 2019-08-17
Create a custom placeholder cluster if system does not know the cluster

cdjackson:
This adds a simple class ```ZclCustomCluster``` as a placeholder for any clusters that are unknown by the system. This ensures that the cluster will appear in the cluster lists for the endpoint and attributes may be read and written.

The system allows updating of the default cluster through the ```addInputCluster``` and ```addOutputCluster``` methods to allow the application to add a specific implementation.

@triller-telekom you might want to take a look at this to ensure it does not impact your implementation of manufacturer specific clusters - I don't think it should, but I welcome any feedback before I merge.


cdjackson:
https://github.com/openhab/org.openhab.binding.zigbee/issues/429

cdjackson:
#710

Issue #712, opened 2019-08-13
Centralite plug turns off and on by itself


TomTravaglino:
We're experiencing an issue with a Centralite 3200 Smart Outlet. After pairing the device and having it turned on we can observe the following behavior: About every 5 minutes the plug turns off by itself and after about 1.5 seconds it turns on again. In the sniffer logs we see a `ZCL: ReportAttributes` when the plug turns on but nothing when it turns off. We also see that this always occurs in combination with these commands:
![Screenshot from 2019-08-12 15-01-17](https://user-images.githubusercontent.com/16982312/62915869-562c5200-bd96-11e9-879c-5a0c5976dbaa.png)

And there is another anomaly in the logs: Every 2 seconds the plug broadcasts a `IEEE Address Request` and does not get a response.

We can reliably reproduce all this on two copies.

I attached a zip containing runtime logs and the sniffer logs produced by Simplicity Studio. The network key for the sniffer logs is `C42FA032417A93E4EC6EA460302B35D0`.

[logs.zip](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/files/3495394/logs.zip)

TomTravaglino:
@cdjackson Friendly reminder. Do you have an idea what's wrong with the device?

cdjackson:
Sorry - I’ve not really looked at this as it seems to be an issue with the device from your description.

cdjackson:
> every 5 minutes the plug turns off by itself and after about 1.5 seconds it turns on again.

The only thing I can think of here is that this is a feature of the device, possibly in relation to how the library is handling some low level acks (I think I mentioned this to @triller-telekom in our last call). I might be wrong though as it would be non-standard and therefore the only way to know is from the device manufacturer.

> And there is another anomaly in the logs: Every 2 seconds the plug broadcasts a IEEE Address Request and does not get a response.

What response are you expecting? Devices are allowed to send such requests - the framework does not have to reply and in this case the request does not appear valid from my initial look.

cdjackson:
Closing this due to inactivity. If the issue persists, and you think it's related to the framework, then please provide further information.

PR #734, opened 2019-09-01
Improve ZclCommand processing and handling of DefaultResponse

cdjackson:
This PR resolves a major non-compliance of the framework to the ZigBee standards in relation to the ```DefaultResponse```. It does present some small breaking changes in the way that commands are processed in ```ZclCluster``` and ```ZigBeeApplication``` in order to get consistent processing of ```DefaultResponse``` frames.

**ZigBee Requirement from ZCL Specification**

The Default Response command SHALL be generated when all 4 of these criteria are met: 
1. A device receives a unicast command that is not a Default Response command. 
2. No other command is sent in response to the received command, using the same Transaction sequence number as the received command. 
3. The Disable Default Response bit of its Frame control field is set to 0 (see 2.4.1.1.4) or when an error results. 
4. The “Effect on Receipt” clause for the received command does not override the behavior of when a Default Response command is sent. 

If a device receives a command in error through a broadcast or multicast transmission, the command SHALL be discarded and the Default Response command SHALL not be generated.

If the identifier of the received command is not supported on the device, it SHALL set the command identifier field to the value of the identifier of the command received in error. The status code field SHALL be set to the either ```UNSUP_CLUSTER_COMMAND```, ```UNSUP_GENERAL_COMMAND```, ```UNSUP_MANUF_CLUSTER_COMMAND``` or ```UNSUP_MANUF_GENERAL_COMMAND```, as appropriate.
If the device receives a unicast cluster command to a particular endpoint, and the cluster does not exist on the endpoint, the status code field SHALL be set to ```UNSUPPORTED_CLUSTER```.
The Default Response command SHALL be generated in response to reception of all commands, including response commands (such as the Write Attributes Response command), under the conditions specified above. However, the Default Response command SHALL not be generated in response to reception of another Default Response command. 

**Framework Response Processing**

Each ZCL command that is processed in the framework must be responded to within the command handler thread by returning either true if the handler has, or will, send a response, or false if it has not processed the command. If all handlers return false, the ```ZclCluster``` handler will respond with a ```DefaultResponse``` with an UNSUPPORTED error. Note that handlers do not need to check if the Disable default Response flag is set - the ```ZclCluster``` handler will manage this.

**Implementation**

Commands are sent from ```ZigBeeNetworkManager``` to the ```ZigBeeNode```, then to the ```ZigBeeEndpoint``` and on to the ```ZclCluster```.

* ```ZigBeeNode``` will send a ```DefaultResponse``` with ```UNSUPPORTED_CLUSTER``` if the endpoint is not supported.
* ```ZigBeeEndpoint``` will send a ```DefaultResponse``` with ```UNSUPPORTED_CLUSTER``` if the cluster is not supported.
* ```ZclCluster``` will send a ```DefaultResponse``` with ```UNSUP_CLUSTER_COMMAND```, ```UNSUP_GENERAL_COMMAND```, ```UNSUP_MANUF_CLUSTER_COMMAND``` or ```UNSUP_MANUF_GENERAL_COMMAND```, as appropriate if there are no listeners registered, if the command is not processed within a set time, or no listeners send a response (ie return ```true```).

When ```ZclCluster``` calls each ```ZclCommandListener``` it waits for each handler to complete. Once all handlers have completed, if none of them sent a response (as determined by the handler returning ```true```) the above action is taken to send a ```DefaultResponse```.

The static helper method ```ZclCluster.createDefaultResponse()``` is added to allow a consistent response from different parts of the framework (ie the ```ZclNode``` and ```ZclEndpoint```.

To ensure that all cluster commands are handled consistently with respect to the ```DefaultResponse```, the application command handler ```ZigBeeApplication. commandReceived ()``` has been removed. Implementations of  ```ZigBeeApplication``` should register to receive commands via the command listeners in the respective clusters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,8 @@ Currently implemented extensions include -:
 
 These provide basic functionality and can be extended as required to meet the application needs.
 
+It should be noted that extensions to the framework should be performed by linking to the ```ZclCluster``` through the ```ZclCommandListener``` notifications. This ensures that processing of ```DefaultResponse``` commands is handled correctly as the framework will automatically handle sending of these responses if there is no response sent. Responses to incoming commands must be returned from the ```handleCommand``` method to ensure that the framework correctly handles the ```DefaultResponse``` notifications. If not handler returns a response, then the framework will return a ```DefaultResponse``` with ```UNSUP_CLUSTER_COMMAND```, ```UNSUP_GENERAL_COMMAND```, ```UNSUP_MANUF_CLUSTER_COMMAND``` or ```UNSUP_MANUF_GENERAL_COMMAND```, as appropriate.
+
 ## Overview
 
 Extensions may provide different levels of functionality - an extension may be as simple as configuring the framework to work with a specific feature, or could provide a detailed application.
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Currently implemented extensions include -:
 
 These provide basic functionality and can be extended as required to meet the application needs.
 
-It should be noted that extensions to the framework should be performed by linking to the ```ZclCluster``` through the ```ZclCommandListener``` notifications. This ensures that processing of ```DefaultResponse``` commands is handled correctly as the framework will automatically handle sending of these responses if there is no response sent. Responses to incoming commands must be returned from the ```handleCommand``` method to ensure that the framework correctly handles the ```DefaultResponse``` notifications. If not handler returns a response, then the framework will return a ```DefaultResponse``` with ```UNSUP_CLUSTER_COMMAND```, ```UNSUP_GENERAL_COMMAND```, ```UNSUP_MANUF_CLUSTER_COMMAND``` or ```UNSUP_MANUF_GENERAL_COMMAND```, as appropriate.
+It should be noted that extensions to the framework should be performed by linking to the ```ZclCluster``` through the ```ZclCommandListener``` notifications. This ensures that processing of ```DefaultResponse``` commands is handled correctly as the framework will automatically handle sending of these responses if there is no response sent. Calls to the ```handleCommand``` method must return ```true``` if the method has sent a response, or ```false``` if no response is sent to ensure that the framework correctly handles the ```DefaultResponse``` notifications. If all handlers return a ```false``` response, then the framework will send a ```DefaultResponse``` with ```UNSUP_CLUSTER_COMMAND```, ```UNSUP_GENERAL_COMMAND```, ```UNSUP_MANUF_CLUSTER_COMMAND``` or ```UNSUP_MANUF_GENERAL_COMMAND```, as appropriate.
 
 ## Overview
 
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
@@ -18,10 +18,13 @@ import java.util.TreeSet;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.RunnableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -635,7 +638,9 @@ public abstract class ZclCluster {
      * @param status the {@link ZclStatus} to send in the response
      * @param manufacturerCode the manufacturer code to set in the response (or null, if the command is not
      *            manufacturer-specific, or if the cluster is itself manufacturer-specific)
+     * @deprecated use sendDefaultResponse(ZclCommand command, ZclStatus status)
      */
+    @Deprecated
     public void sendDefaultResponse(Integer transactionId, Integer commandIdentifier, ZclStatus status,
             Integer manufacturerCode) {
         DefaultResponse defaultResponse = new DefaultResponse();
@@ -651,7 +656,7 @@ public abstract class ZclCluster {
             defaultResponse.setManufacturerCode(manufacturerCode);
         }
 
-        zigbeeEndpoint.sendTransaction(defaultResponse);
+        send(defaultResponse);
     }
 
     /**
@@ -660,9 +665,30 @@ public abstract class ZclCluster {
      * @param transactionId the transaction ID to use in the response
      * @param commandIdentifier the command identifier to which this is a response
      * @param status the {@link ZclStatus} to send in the response
+     * @deprecated use sendDefaultResponse(ZclCommand command, ZclStatus status)
      */
+    @Deprecated
     public void sendDefaultResponse(Integer transactionId, Integer commandIdentifier, ZclStatus status) {
         sendDefaultResponse(transactionId, commandIdentifier, status, null);
+    }
+
+    /**
+     * Sends a default response to the client
+     *
+     * @param command the {@link ZclCommand} to which we are responding
+     * @param status the {@link ZclStatus} to send in the response
+     */
+    public void sendDefaultResponse(ZclCommand command, ZclStatus status) {
+        ZclCommand response = createDefaultResponse(command, status);
+        if (response == null) {
+            return;
+        }
+        if (isManufacturerSpecific()) {
+            response.setManufacturerCode(getManufacturerCode());
+        }
+
+        // Default response will not solicit a response
+        zigbeeEndpoint.sendTransaction(response);
     }
 
     /**
@@ -1017,12 +1043,6 @@ public abstract class ZclCluster {
      * @param listener the {@link ZclAttributeListener} to add
      */
     public void addAttributeListener(ZclAttributeListener listener) {
-        // Don't add more than once.
-        if (attributeListeners.contains(listener)) {
-            logger.trace("{}: ZclCluster.addAttributeListener already contains {}", zigbeeEndpoint.getEndpointAddress(),
-                    listener);
-            return;
-        }
         logger.trace("{}: ZclCluster.addAttributeListener adding {}", zigbeeEndpoint.getEndpointAddress(), listener);
         attributeListeners.add(listener);
     }
@@ -1065,10 +1085,6 @@ public abstract class ZclCluster {
      */
     public void addCommandListener(ZclCommandListener listener) {
         logger.trace("{}: ZclCluster.addCommandListener({})", zigbeeEndpoint.getEndpointAddress(), listener);
-        // Don't add more than once.
-        if (commandListeners.contains(listener)) {
-            return;
-        }
         commandListeners.add(listener);
     }
 
@@ -1086,8 +1102,11 @@ public abstract class ZclCluster {
      * Notify command listeners of an received {@link ZclCommand}.
      *
      * @param command the {@link ZclCommand} to notify
+     * @return true if at least one of the command handlers has responded to the command
      */
-    private void notifyCommandListener(final ZclCommand command) {
+    private boolean notifyCommandListener(final ZclCommand command) {
+        final AtomicBoolean response = new AtomicBoolean();
+        CountDownLatch latch = new CountDownLatch(commandListeners.size());
         for (final ZclCommandListener listener : commandListeners) {
             NotificationService.execute(new Runnable() {
                 @Override
@@ -1095,9 +1114,27 @@ public abstract class ZclCluster {
                     logger.trace("{}: ZclCluster.notifyCommandListener {} of {}", zigbeeEndpoint.getEndpointAddress(),
                             listener, command);
 
-                    listener.commandReceived(command);
+                    if (listener.commandReceived(command)) {
+                        // A thread is responding - we don't need to wait around
+                        for (long cnt = latch.getCount(); cnt > 0; cnt--) {
+                            latch.countDown();
+                            response.set(true);
+                        }
+                        latch.notifyAll();
+                    }
+                    latch.countDown();
                 }
             });
+        }
+
+        try
+
+        {
+            // TODO: Set the timer properly
+            latch.await(1, TimeUnit.SECONDS);
+            return response.get();
+        } catch (InterruptedException e) {
+            return false;
         }
     }
 
@@ -1112,6 +1149,8 @@ public abstract class ZclCluster {
         for (AttributeReport report : command.getReports()) {
             updateAttribute(report.getAttributeIdentifier(), report.getAttributeValue());
         }
+
+        sendDefaultResponse(command, ZclStatus.SUCCESS);
     }
 
     /**
@@ -1132,6 +1171,8 @@ public abstract class ZclCluster {
 
             updateAttribute(record.getAttributeIdentifier(), record.getAttributeValue());
         }
+
+        sendDefaultResponse(command, ZclStatus.SUCCESS);
     }
 
     private void updateAttribute(int attributeId, Object attributeValue) {
@@ -1170,8 +1211,28 @@ public abstract class ZclCluster {
 
         // If this is a specific cluster command, pass the command to the cluster command handler
         if (!command.isGenericCommand()) {
-            notifyCommandListener(command);
+            if (notifyCommandListener(command)) {
+                return;
+            }
         }
+
+        ZclStatus responseStatus;
+
+        if (command.isManufacturerSpecific()) {
+            if (command.isGenericCommand()) {
+                responseStatus = ZclStatus.UNSUP_MANUF_GENERAL_COMMAND;
+            } else {
+                responseStatus = ZclStatus.UNSUP_MANUF_CLUSTER_COMMAND;
+            }
+        } else {
+            if (command.isGenericCommand()) {
+                responseStatus = ZclStatus.UNSUP_GENERAL_COMMAND;
+            } else {
+                responseStatus = ZclStatus.UNSUP_CLUSTER_COMMAND;
+            }
+        }
+
+        sendDefaultResponse(command, responseStatus);
     }
 
     /**
@@ -1551,6 +1612,32 @@ public abstract class ZclCluster {
             }
         }
         return true;
+    }
+
+    /**
+     * Generates a {@link DefaultResponse} to the supplied {@link ZclCommand} with the requested {@link ZclStatus} code.
+     * If the {@link ZclCommand} does not require a response, this method will return null.
+     *
+     * @param command the received {@link ZclCommand} to which the response is being generated
+     * @param status the {@link ZclStatus} to return in the response
+     * @return the {@link DefaultResponse} ready to send or null if it is not required
+     */
+    public static DefaultResponse createDefaultResponse(ZclCommand command, ZclStatus status) {
+        if (command.isDisableDefaultResponse()) {
+            return null;
+        }
+        DefaultResponse defaultResponse = new DefaultResponse();
+        defaultResponse.setTransactionId(command.getTransactionId());
+        defaultResponse.setCommandIdentifier(command.getCommandId());
+        defaultResponse.setDestinationAddress(command.getDestinationAddress());
+        defaultResponse.setClusterId(command.getClusterId());
+        defaultResponse.setStatusCode(status);
+
+        if (command.isManufacturerSpecific()) {
+            defaultResponse.setManufacturerCode(command.getManufacturerCode());
+        }
+
+        return defaultResponse;
     }
 
 }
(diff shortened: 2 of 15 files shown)


cdjackson:
@triller-telekom you might want to take a look at this for a couple of reasons -:

1) General review please :)
2) The ```sendDefaultResponse``` methods have changed and the manufacturer specific version should be checked.
3) It's possible (but I'm speculating!) that this is also related to #712 

I still need to do more testing on this, but any comments welcome.

README.md
```diff
@@ -203,6 +203,8 @@ Currently implemented extensions include -:
 
 These provide basic functionality and can be extended as required to meet the application needs.
 
+It should be noted that extensions to the framework should be performed by linking to the ```ZclCluster``` through the ```ZclCommandListener``` notifications. This ensures that processing of ```DefaultResponse``` commands is handled correctly as the framework will automatically handle sending of these responses if there is no response sent. Responses to incoming commands must be returned from the ```handleCommand``` method to ensure that the framework correctly handles the ```DefaultResponse``` notifications. If not handler returns a response, then the framework will return a ```DefaultResponse``` with ```UNSUP_CLUSTER_COMMAND```, ```UNSUP_GENERAL_COMMAND```, ```UNSUP_MANUF_CLUSTER_COMMAND``` or ```UNSUP_MANUF_GENERAL_COMMAND```, as appropriate.
```
triller-telekom: typo: If no~~t~~ handler returns a response

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -660,11 +665,32 @@ public void sendDefaultResponse(Integer transactionId, Integer commandIdentifier
...
+     * @param status the {@link ZclStatus} to send in the response
+     */
+    public void sendDefaultResponse(ZclCommand command, ZclStatus status) {
+        ZclCommand response = createDefaultResponse(command, status);
+        if (response == null) {
+            return;
+        }
+        if (isManufacturerSpecific()) {
+            response.setManufacturerCode(getManufacturerCode());
+        }
```
triller-telekom: I the now deprecated method there is also an `else` branch. Are you setting the `manufacturerCode` value somewhere else now?

```
if (isManufacturerSpecific()) {
            defaultResponse.setManufacturerCode(getManufacturerCode());
        } else if (manufacturerCode != null) {
            defaultResponse.setManufacturerCode(manufacturerCode);
        }
```

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -1017,12 +1043,6 @@ public Boolean call() throws Exception {
      * @param listener the {@link ZclAttributeListener} to add
      */
     public void addAttributeListener(ZclAttributeListener listener) {
-        // Don't add more than once.
-        if (attributeListeners.contains(listener)) {
```
triller-telekom: If I recall this correctly: you have just introduced that check to avoid duplicates. What is the reason to remove this check again? Anyway its a `CopyOnWriteArraySet` so I wonder how duplicates can happen at all here.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -1065,10 +1085,6 @@ public void run() {
      */
     public void addCommandListener(ZclCommandListener listener) {
         logger.trace("{}: ZclCluster.addCommandListener({})", zigbeeEndpoint.getEndpointAddress(), listener);
-        // Don't add more than once.
-        if (commandListeners.contains(listener)) {
```
triller-telekom: See above.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -1086,19 +1102,40 @@ public void removeCommandListener(final ZclCommandListener listener) {
...
+                    }
+                    latch.countDown();
                 }
             });
         }
+
+        try
+
+        {
+            // TODO: Set the timer properly
```
triller-telekom: Out of curiosity: How do you plan to estimate a proper waiting time here?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCommand.java
```diff
@@ -122,10 +122,37 @@ public Integer getManufacturerCode() {
...
+     */
+    public void setDisableDefaultResponse(boolean disableDefaultResponse) {
+        this.disableDefaultResponse = disableDefaultResponse;
+    }
+
     @Override
     public String toString() {
         Integer resolvedClusterId = getClusterId();
-        final StringBuilder builder = new StringBuilder();
+        final StringBuilder builder = new StringBuilder(50);
```
triller-telekom: Whats the reason for artificially starting with a length of 50 here?

cdjackson:
Thanks Stefan,
I did some testing last night and it looked ok. It did reveal a bug which I need to look into further - it might be as a result of this PR or it might be something that is already in the framework and just exposed by this PR. The sending of the default responses looked ok though at first glance and once I’ve tested some more I’ll let you know.

Cheers
Chris

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
cdjackson: Yes - in the ```createDefaultResponse``` it handles the check if the command has the manufacturer specific flag set. This cluster specific check can't be done in that method as it's a static helper method that's also called from the ```ZigBeeNode``` and ```ZigBeeEndpoint```.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
cdjackson: I think you are getting muddled with another PR recently for a different listener - possibly in the command notification class.

Here ```attributeListeners``` is a ```Set``` so it should not be required to check for duplicates right?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
cdjackson: I plan to just set it statically (ie provide a method to allow it to be set, but not do anything to estimate it dynamically).

I'm open to other ideas though ;)

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCommand.java
cdjackson: It seemed like a reasonably number to cover the label etc. Otherwise I think it defaults to 16 and will most likely require a reallocation almost immediately. Maybe it could be reduced to 30 or 40 if you think it's too much?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
cdjackson: As above.

README.md
cdjackson: Fixed.

cdjackson:
@triller-telekom From my initial testing this PR looks ok -:

```
22:23:57.261  DEBUG  RX APS: ZigBeeApsFrame [sourceAddress=25855/1, destinationAddress=0/1, profile=0104, cluster=0006, addressMode=null, radius=0, apsSecurity=false, apsCounter=A0, payload=08 06 0A 00 00 10 00]
22:23:57.261  DEBUG  RX ZCL: ZclHeader [frameType=ENTIRE_PROFILE_COMMAND, manufacturerSpecific=false, direction=SERVER_TO_CLIENT, disableDefaultResponse=false, manufacturerCode=0, sequenceNumber=6, commandId=10]
22:23:57.261  DEBUG  RX CMD: ReportAttributesCommand [On/Off: 25855/1 -> 0/1, cluster=0006, TID=06, reports=[Attribute Report: attributeDataType=BOOLEAN, attributeIdentifier=0, attributeValue=false]]
22:23:57.262  DEBUG  000D6FFFFECD75F1: Added transaction to queue, len=1, transaction=ZigBeeTransaction [queueTime=0, state=WAITING, sendCnt=0, command=DefaultResponse [On/Off: 0/0 -> 25855/1, cluster=0006, TID=06, commandIdentifier=10, statusCode=SUCCESS]]
22:23:57.262  DEBUG  25855/1: Sending ZigBeeTransaction [queueTime=0, state=WAITING, sendCnt=0, command=DefaultResponse [On/Off: 0/0 -> 25855/1, cluster=0006, TID=06, commandIdentifier=10, statusCode=SUCCESS]]
22:23:57.262  DEBUG  TX CMD: DefaultResponse [On/Off: 0/0 -> 25855/1, cluster=0006, TID=06, commandIdentifier=10, statusCode=SUCCESS]
22:23:57.262  DEBUG  TX ZCL: ZclHeader [frameType=ENTIRE_PROFILE_COMMAND, manufacturerSpecific=false, direction=CLIENT_TO_SERVER, disableDefaultResponse=false, manufacturerCode=0, sequenceNumber=6, commandId=11]
22:23:57.262  DEBUG  TX APS: ZigBeeApsFrame [sourceAddress=0/1, destinationAddress=25855/1, profile=0104, cluster=0006, addressMode=DEVICE, radius=31, apsSecurity=false, apsCounter=7D, payload=00 06 0B 0A 00]
```

The above shows the ```DefaultResponse``` being returned from the framework in the response to an attribute report.  I've not tested other responses as it's not super simple, but there are a reasonable set of tests in this PR to exercise this.

Note that there will be a PR coming shortly to address #735 which really needs to be resolved before the ```DefaultResponse``` changes here will work properly, otherwise there will be multiple responses sent (most of them being ```UNSUPPORTED_CLUSTER``` responses - so this is also tested ;) ).

cdjackson:
This is now rebased with the node command listener fix.

triller-telekom:
I am currently experimenting with this PR and what I noticed is that ZSS is responding twice to `QueryNextImageCommand`s. There are two `DefaultResponse`s, both with the same `Tid=61`.

```
2019-09-04T14:59:05.123+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler] - <-- RX SPI frame: FE 25 25 88 FF 00 45 00 04 01 19 00 04 01 00 01 00 00 90 FF EB AB 50 FF FF 0C 01 61 01 00 94 27 01 04 2A 00 00 00 02 
2019-09-04T14:59:05.123+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler] - RX EZSP: EzspIncomingMessageHandler [type=EMBER_INCOMING_UNICAST, apsFrame=EmberApsFrame [profileId=260, clusterId=25, sourceEndpoint=4, destinationEndpoint=1, options=[EMBER_APS_OPTION_ENABLE_ROUTE_DISCOVERY], groupId=0, sequence=90], lastHopLqi=255, lastHopRssi=-21, sender=20651, bindingIndex=255, addressIndex=255, messageContents=01 61 01 00 94 27 01 04 2A 00 00 00]
2019-09-04T14:59:05.124+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp] - RX EZSP: EzspIncomingMessageHandler [type=EMBER_INCOMING_UNICAST, apsFrame=EmberApsFrame [profileId=260, clusterId=25, sourceEndpoint=4, destinationEndpoint=1, options=[EMBER_APS_OPTION_ENABLE_ROUTE_DISCOVERY], groupId=0, sequence=90], lastHopLqi=255, lastHopRssi=-21, sender=20651, bindingIndex=255, addressIndex=255, messageContents=01 61 01 00 94 27 01 04 2A 00 00 00]
2019-09-04T14:59:05.125+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - RX APS: ZigBeeApsFrame [sourceAddress=20651/4, destinationAddress=0/1, profile=0104, cluster=0019, addressMode=null, radius=0, apsSecurity=false, apsCounter=90, payload=01 61 01 00 94 27 01 04 2A 00 00 00]
2019-09-04T14:59:05.125+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - RX ZCL: ZclHeader [frameType=CLUSTER_SPECIFIC_COMMAND, manufacturerSpecific=false, direction=CLIENT_TO_SERVER, disableDefaultResponse=false, manufacturerCode=0, sequenceNumber=97, commandId=1]
2019-09-04T14:59:05.126+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - RX CMD: QueryNextImageCommand [Ota Upgrade: 20651/4 -> 0/1, cluster=0019, TID=61, fieldControl=0, manufacturerCode=10132, imageType=1025, fileVersion=42, hardwareVersion=null]
2019-09-04T14:59:05.126+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.transaction.ZigBeeTransactionManager] - notifyTransactionCommand: QueryNextImageCommand [Ota Upgrade: 20651/4 -> 0/1, cluster=0019, TID=61, fieldControl=0, manufacturerCode=10132, imageType=1025, fileVersion=42, hardwareVersion=null] 
2019-09-04T14:59:05.126+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.app.otaserver.ZclOtaUpgradeServer] - 20651/4 OTA Error: Invalid server state OTA_UNINITIALISED when handling QueryNextImageCommand.
2019-09-04T14:59:05.127+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.transaction.ZigBeeTransactionQueue] - 00124B000253FB4A: Added transaction to queue, len=1, transaction=ZigBeeTransaction [queueTime=0, state=WAITING, sendCnt=0, command=DefaultResponse [Ota Upgrade: 0/0 -> 20651/4, cluster=0019, TID=61, commandIdentifier=1, statusCode=UNKNOWN]]
2019-09-04T14:59:05.128+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.transaction.ZigBeeTransactionManager] - 20651/4: Sending ZigBeeTransaction [queueTime=0, state=WAITING, sendCnt=0, command=DefaultResponse [Ota Upgrade: 0/0 -> 20651/4, cluster=0019, TID=61, commandIdentifier=1, statusCode=UNKNOWN]]
2019-09-04T14:59:05.128+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.transaction.ZigBeeTransactionManager] - transactionListenerAdded: 1 outstanding
2019-09-04T14:59:05.128+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - TX CMD: DefaultResponse [Ota Upgrade: 0/0 -> 20651/4, cluster=0019, TID=61, commandIdentifier=1, statusCode=UNKNOWN]
2019-09-04T14:59:05.128+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - TX ZCL: ZclHeader [frameType=ENTIRE_PROFILE_COMMAND, manufacturerSpecific=false, direction=CLIENT_TO_SERVER, disableDefaultResponse=false, manufacturerCode=0, sequenceNumber=97, commandId=11]
2019-09-04T14:59:05.129+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.ZigBeeNetworkManager] - TX APS: ZigBeeApsFrame [sourceAddress=0/1, destinationAddress=20651/4, profile=0104, cluster=0019, addressMode=DEVICE, radius=31, apsSecurity=false, apsCounter=24, payload=00 61 0B 01 FF]
2019-09-04T14:59:05.130+0200   445  DEBUG          [symbolicName=com.zsmartsystems.zigbee.dongle.ember, className=com.zsmartsystems.zigbee.dongle.ember.internal.spi.SpiFrameHandler] - TX EZSP queue: 1
2019-09-04T14:59:05.130+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.app.otaserver.ZclOtaUpgradeServer] - 20651/4 OTA Error: Invalid server state OTA_UNINITIALISED when handling QueryNextImageCommand.
2019-09-04T14:59:05.131+0200   432  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.transaction.ZigBeeTransactionQueue] - 00124B000253FB4A: Added transaction to queue, len=1, transaction=ZigBeeTransaction [queueTime=0, state=WAITING, sendCnt=0, command=DefaultResponse [Ota Upgrade: 0/0 -> 20651/4, cluster=0019, TID=61, commandIdentifier=1, statusCode=UNKNOWN]]
```

If I look at this in the sniffer I see these 2 responses, both having the same `Sequence number` as in the request:

double default response

cdjackson:
I don't think this issue you raise is related to this PR. I can't see any way that this can happen unless the command listener is registered twice, but that doesn't seem to be the case. I did find that the command listener is not unregistered when the server closes so I've fixed that, but it's not really a significant issue.

There are tests to check that the ```ZclCluster.handleCommand()``` will only be called once from the ```ZigBeeEndpoint.commandReceived()```. So, all I can think of is that the handler is registered twice or something? The handler actually can't be registered twice as it's a ```Set``` but maybe there are two ```ZclOtaUpgradeServer```s instantiated?

Are you using this full PR exactly?

triller-telekom:
Yes i am using this PR fully. I just made sure to also restart everything which is using the ZSS library and I still see those doubled responses.

However, they seem to ONLY occur as a response to `QueryNextImageCommand`s. Other commands seem to be not affected.

cdjackson:
Are you able to debug this? If not, please provide a full log from startup as I can't see any problem given what you provided yesterday.

cdjackson:
@triller-telekom I need to get this merged over the next couple of days. If you have any further feedback, please let me know.

Thanks.

triller-telekom:
I wish I could provide more feedback here, however this seems to be reproducible only if there is more than one device active in the network.

I used a bitron 902010/28 smart plug and there was only one `DefaultResponse` per `QueryNextImageCommand`. If I turned on some more paired devices like a 902010/25 smart plug and a centralite 3200D smart plug I saw that those `DefaultResponses` were send twice.

IMHO this is not a defective device because once I have turned on a couple of devices they receive also doubled `DefaultResponse`s as a reply to a `QueryNextImageCommand`.

So its quite hard to debug and since we are moving offices all the equipment is packed now, etc. :(

Nevertheless I was able to pair devices, control devices and even pair a Centralite 3310 which uses manufacturer specific attributes.

So in general this PR shouldn't break things and if you are in a hurry to merge this PR, then do it and we can track the doubled `DefaultResponse`s in a separate issue.

cdjackson:
Thanks Stefan,
The fact that it works fine with only one device is useful. I will see if I can spot anything wrong in that sort of area, although as all commands should come through the clusters, they should be filtered for address etc.

Cheers
Chris

cdjackson:
@triller-telekom one possible reason for your issue could be in the ```ZigBeeOtaUpgradeExtension``` class where the ```ZclOtaUpgradeServer``` is created. Here I could imagine that it is possible if ```nodeAdded``` gets called twice that two separate ```ZclOtaUpgradeServer```s could be instantiated for the same device. This would only happen if the ```nodeAdded``` callback was made after the clusters are added and I wouldn't have expected this in general (although it is definitely possible).

I think I will open an issue on this - it's about the only thing that I can see that can cause what you are seeing - I welcome your thoughts though :)

```
    @Override
        for (ZigBeeEndpoint endpoint : node.getEndpoints()) {
            if (endpoint.getOutputCluster(ZclOtaUpgradeCluster.CLUSTER_ID) != null) {
                endpoint.addApplication(new ZclOtaUpgradeServer());
                break;
            }
        }
    }
```

Issue #737, opened 2019-09-03
Telegesis dongle: Transaction retry counter not reset


yoadey:
**Describe the bug**
In TelegesisFrameHandler exists a retry counter for failed transactions. After a successful transaction it should be reset to zero, but currently the reset is not implemented. This leads to a stop of the communication with the dongle if there occur 3 timeouts in total and not just directly after each other.
In log, the message `TELEGESIS Timer: Timeout #` counts up to 3 while many other commands have been successful send.

**To Reproduce**
1. Start zigbee with a telegesis dongle
2. Send many commands, even if they have not been fully processed yet

**Expected behavior**
All commands are processed or dropped, communication with the dongle stays alive.

**Additional context**
Is in context with https://github.com/openhab/org.openhab.binding.zigbee/issues/474, but does not fully solve this issue.

Will soon add a pull request for this

cdjackson:
Closed by #738

PR #738, opened 2019-09-03
Telegesis: retries of transaction timeout were not reset

yoadey:
References issue #737

Retry counter of transaction timeout for telegesis dongle is not reset.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.telegesis/src/test/java/com/zsmartsystems/zigbee/dongle/telegesis/internal/TelegesisFrameHandlerTest.java
+++ b/com.zsmartsystems.zigbee.dongle.telegesis/src/test/java/com/zsmartsystems/zigbee/dongle/telegesis/internal/TelegesisFrameHandlerTest.java
@@ -145,9 +145,9 @@ public class TelegesisFrameHandlerTest {
         // Repeat the task four times, with timeouts after every second. Retries should be reset after each round
         for (int i = 0; i < 3; i++) {
             respond[0] = false;
-            sendCommandAndWait(frameHandler, processed);
+            sendCommandAndWait(frameHandler, processed, respond);
             respond[0] = true;
-            sendCommandAndWait(frameHandler, processed);
+            sendCommandAndWait(frameHandler, processed, respond);
         }
 
         Mockito.verify(dongle, Mockito.times(0)).notifyStateUpdate(false);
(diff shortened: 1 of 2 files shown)


cdjackson:
@yoadey thanks for this - I will take a look at it shortly. Please can you sign the CLA as per the above comment.

Thanks.

yoadey:
I have already signed it and CLA signed test is green for me, is it still orange?

cdjackson:
Sorry - it looks fine now, thanks. I think I looked at it, then came back a little later and replied, so you probably got there in-between.

Thanks.

com.zsmartsystems.zigbee.dongle.telegesis/src/test/java/com/zsmartsystems/zigbee/dongle/telegesis/internal/TelegesisFrameHandlerTest.java
```diff
@@ -168,6 +248,33 @@ public void run() {
...
+        command.setRegister(0);
+        command.setPassword("password");
+        command.setState(false);
+
+        waitTime[0] = time;
+        lastSendCommandTime[0] = System.currentTimeMillis();
+        processed[0] = false;
+        frameHandler.queueFrame(command);
+        long timeout = System.currentTimeMillis();
+        while (!processed[0] && System.currentTimeMillis() - timeout < 200) {
```
cdjackson: I'm not sure I'm keen on this - such a short timeout (200ms) will likely cause problems with CI when running on CI servers running multiple builds (as we have). This will probably cause occasional failures.

Is there a need to have such a short delay?

I'm also wondering if this helper method is really needed or if we can just use ```Mockito.verify``` to wait for ``` the response?

cdjackson:
https://github.com/openhab/org.openhab.binding.zigbee/pull/477
https://github.com/openhab/org.openhab.binding.zigbee/issues/474

cdjackson:
@yoadey I just wanted to check if you'd seen my comments?  I want to do a binding update of the ZSS libs at some point in the next week or so and it would be good to get this into the libraries for that update.

yoadey:
@cdjackson yes, I've seen your comments, but was very busy the last week and didn't come up with a good solution to which the timeout should be set. After all, if we just increase the timeout the test runs much longer which I don't like. I'll try to have an updated test till tomorrow evening, is that OK?

cdjackson:
No problem - I didn’t mean to push you if you’re busy but wasn’t sure if you’d spotted the comments. 

Thanks

yoadey:
Have updated the tests, timeouts are increased heavily, so there shouldn't occur any problems on slower machines and made use of internal field sentCommand to improve tests speed. Also cleaned the tests a little bit up

cdjackson:
Thanks.

PR #742, opened 2019-09-07
Add checks to ZigBeeOtaUpgradeExtension to not duplicate ZclOtaUpgradeServer

### LINKED ISSUE #740: ZclOtaUpgradeServer may be instantiated multiple times per node
As reported by @triller-telekom in #734, it is possible for the ```ZclOtaUpgradeServer``` to send multiple ```DefaultResponse``` commands.

One possible reason for this issue could be in the ```ZigBeeOtaUpgradeExtension``` class where the ```ZclOtaUpgradeServer``` is created. Here I could imagine that it is possible if ```nodeAdded``` gets called twice that two separate ```ZclOtaUpgradeServer```s could be instantiated for the same device. This would only happen if the ```nodeAdded``` callback was made after the clusters are added and I wouldn't have expected this in general (although it is definitely possible).

```
    @Override
        for (ZigBeeEndpoint endpoint : node.getEndpoints()) {
            if (endpoint.getOutputCluster(ZclOtaUpgradeCluster.CLUSTER_ID) != null) {
                endpoint.addApplication(new ZclOtaUpgradeServer());
                break;
            }
        }
    }
```

The ```nodeAdded``` will be called as endpoints are added. In general, it should only be called once by the standard discovery service, although other discovery services used by alternative systems could cause this to be called multiple times.

It therefore seems pertinent to protect against this issue. 

Looking at the implementation in ```ZigBeeEndpoint``` there is in fact another problem in that only the second server will be retained -:

```
    public void addApplication(ZigBeeApplication application) {
        applications.put(application.getClusterId(), application);
        ZclCluster cluster = outputClusters.get(application.getClusterId());
        if (cluster == null) {
            cluster = inputClusters.get(application.getClusterId());
        }
        application.appStartup(cluster);
    }
```

In the above, the ```applications.put``` will only remember the last application added - there probably should be a check here to ensure that a second application is not added.

cdjackson:
Closes #740 

Requires #741 to be merged first

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZigBeeOtaUpgradeExtension.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZigBeeOtaUpgradeExtension.java
@@ -48,6 +48,23 @@ public class ZigBeeOtaUpgradeExtension implements ZigBeeNetworkExtension, ZigBee
 
     @Override
     public void nodeAdded(ZigBeeNode node) {
+        /*
+         * This method needs to register the ZclOtaUpgradeServer to an endpoint on the node. Only one endpoint should be
+         * registered with the ZclOtaUpgradeServer.
+         * It should be noted that the nodeAdded method can be called more than once if the endpoint or cluster
+         * information is updated. As such we need to take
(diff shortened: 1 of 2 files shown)


cdjackson:
@triller-telekom it would be interesting to see if this has any impact on the OTA issue with multiple DefaultResponse commands being sent in your system.

PR #762, opened 2019-09-26
return NO_IMAGE_AVAILABLE if no ota file has been provided

sesuncedu:
If the server has not been initialized,  ```handleQueryNextImageCommand``` sends a default response with status UNKNOWN (0xff).  

It would seem that ```NO_IMAGE_AVAILABLE``` should be returned in this case.


cdjackson:
Thanks - I agree - this is probably better. Please can you format the code with the formatter, and also modify the logger message so it uses the standard format as per the other messages.

sesuncedu:
Sure - I made the change on my pi, and forgot how much intellij usually does for me.

sesuncedu:
D'oh! I realized that **none*** of the NO_IMAGE_AVAILABLE responses  should be DefaultResponse! They should all be QueryNextImageResponse! 

That explains why there are more retries than I had expected. 

 ```cluster.queryNextImageResponse(NO_IMAGE_AVAILABLE,null,null,null,null);```

cdjackson:
Good catch - I assume you are planning to update this PR?

cdjackson:
@sesuncedu are you still planning to continue with this PR? I'd prefer to try and close them out quickly before moving onto other issues if possible to avoid merge conflicts (as we now have).

Issue #773, opened 2019-10-02
Heisenbug in test case introduced for #770


sesuncedu:
The final test for invocation count for sendTransaction in the test for ```handleCommand``` generally fails.

Mockito sometimes reports a count of 6 instead of 5. Stepping through the relevant code makes the problem go away, as does adding println. 

Replacing the mocked listener with a lambda does  not change the behavior  (unless it prints). 

This seems to suggest some weird timing or concurrency related mockito issue.

The problem has been observed with oracle jdk 8 and 11, under Linux on amd64 and arm, from maven and under intellij, and in the latter case, when the failing test is the only test invoked.

Workaround may be to isolate the last test in a separate test method.

I will try with mockito verboselogging turned on (though that may count as observing for heisenbug purposes) .

sesuncedu:
The root  cause seems to be in ```ZclCommand::notifyListeners```.  

This method seems rather odd:
1. If there are n listeners, of which m return true, the list will be decremented m*n times
2. If all listeners take too long to complete, the method will return false. 
2. If any listener tasks are executing or queued, they will still be dispatched / run to completion.
3. More than one listener can  handle the command. more than one listener will handle the command. Each listener might send a response.

The simplest fix would be to require all listeners to complete without blocking for an extended time, and execute them sequentially on the main thread.  If a listener wishes to handle the command, it can submit a task to an executor service. 

A more complicated approach could use  ExecutorCompletionService;  the listener would need to be split into decision and handling parts, and if the number of execution threads is bounded it is still possible for slow listeners to cause problems even if a queued  listener would be fast. 

The simple approach seems reasonable here.

cdjackson:
What is the actual root cause though?

It is not suitable to make everything run on the main thread - this will cause significant issues. I don’t think there are any issues with this as it currently stands other than the test, and I don’t understand what you are suggesting is actually the issue with the current implementation?

cdjackson:
> The root cause seems to be in ZclCommand::notifyListeners.

Can you also clarify where you think the error is since the ```ZclCommand``` class doesn't have such a method. I assume you probably are referring to ```ZclCluster::notifyCommandListener```?

cdjackson:
Can you also clarify what test fails - or provide a log showing the failure. I can't replicate it here and it's pretty unclear from your report where the test failure is.

sesuncedu:
The failing test is 
```ZclClusterTest::handleCommand```

[...]

PR #780, opened 2019-10-03
Fix for race condition in ZclCluster::notifyCommandListener per  #773

sesuncedu:
Fix for immediate issue in #773
Fix race condition causing incorrect value to be returned in ```ZclCluster::notifyCommandListener```, wherein the latch count was decremented before the response was set.

In the event of a timeout or interrupted exception, the current value of the response is used. This will give the correct result (possibly after a timeout) if the size of the listener set is reduced after the latch has been created, but before the loop has started. 

This PR does not fix other preexisting concurrency issues such as those involving multiple listeners which may return true. 

The 

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
@@ -1129,23 +1129,32 @@ public abstract class ZclCluster {
                             listener, command);
 
                     if (listener.commandReceived(command)) {
-                        // A thread is responding - we don't need to wait around
-                        for (long cnt = latch.getCount(); cnt > 0; cnt--) {
+                        response.set(true);
+                        while( latch.getCount() > 0) {
                             latch.countDown();
-                            response.set(true);
                         }
+                    } else {
+                        latch.countDown();
                     }
-                    latch.countDown();
                 }
             });
         }
 
         try {
             // TODO: Set the timer properly
-            latch.await(1, TimeUnit.SECONDS);
-            return response.get();
+            if(latch.await(1, TimeUnit.SECONDS)) {
+                final boolean b = response.get();
+                logger.trace(" LATCH done - {}" , b);
+                return b;
+            }else {
+                long cnt = latch.getCount();
+                logger.trace("LATCH TIME OUT, cnt = {}" ,cnt);
+                return response.get();
+            }
         } catch (InterruptedException e) {
-            return false;
+            long cnt = latch.getCount();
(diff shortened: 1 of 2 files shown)


cdjackson:
Thanks. Please can you run the formatter on this and I'll take a look.

What do you mean by -:

> This PR does not fix other preexisting concurrency issues such as those involving multiple listeners which may return true.

I don't think there is any concurrency issue if multiple listeners return true. If ANY listener returns true then this will exit. There should not be any concurrency problem with this - or am I missing something.

sesuncedu:
> Thanks. Please can you run the formatter on this and I'll take a look.

Can you let me know what is not being formatted correctly?  I imported the eclipse formatter settings and   "ESH" is active.  Is something not getting imported correctly? 

> > This PR does not fix other preexisting concurrency issues such as those involving multiple listeners which may return true.
> 
> I don't think there is any concurrency issue if multiple listeners return true. If ANY listener returns true then this will exit. There should not be any concurrency problem with this - or am I missing something.

The issues I fixed do address the race cases.  

The concurrency concern referred to is indeed the overall one of multiple listeners sending responses to a command, which I was deeming to be outside the scope of this pull request, since that may be intended behavior. 

I feel a bit icky relying on timeouts to give correct behavior if a listener is removed at an awkward time, but I would also feel icky making a defensive copy of a copy on write set. It's correct, and faster in all but the edge case. 

The timeout was already there  when I got here, occifer.

cdjackson:
> Can you let me know what is not being formatted correctly?

There are a few things - for starters take a look at the if/else statements - the spacing is not correct so it tells me the code is not formatted.

> The issues I fixed do address the race cases.

Sure - I didn't insinuate that it didn't. My point was just that multiple listeners reporting true is in itself not an issue. I know it results in the potential for multiple responses, which I agree isn't nice, but it's not easy to resolve this. Probably there should not be multiple listeners responding to a message or there is likely another issue in the application.

> The concurrency concern referred to is indeed the overall one of multiple listeners sending responses to a command,

Ok, I don't really think that's a concurrency issue, but understood.

> I feel a bit icky relying on timeouts to give correct behavior if a listener is removed at an awkward time

The listeners should in theory be protected, but I'd not considered what happens if a listener is removed while it's waiting. Maybe there's something else needed if that's the case as I agree that's not so nice. If you have a suggestion to resolve this, then please suggest away :)

sesuncedu:
The formatter was not being auto-applied, but also had some stuff not imported 
(import blocks and annotation blank lines).  

 I usually reflexively hit reformat even when I shouldn't, and so I was looking for spurious diffs.

cdjackson:
Perfect - thanks. I’ll take a look at this tomorrow (it’s getting late here now :) ).

PR #789, opened 2019-10-09
Add a osgi baseline goal to check for incorrect bundle and package versions.

sesuncedu:
Use bnd-baseline-maven-plugin to check for packages and bundles that  have incorrect semantic versions. which are required for OSGI.

This commit sets a flag to continue the build even when errors are found. This should be changed once the underlying issue has been addressed.

This would ideally be done before the next release (which should be bundle version 2.0.0 for most artifacts).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,10 @@
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 		<slf4j.version>1.7.2</slf4j.version>
 		<log4j.version>1.2.17</log4j.version>
+		<bnd.baseline.continueOnError>true</bnd.baseline.continueOnError>
+		<bnd.baseline.failOnMissing>false</bnd.baseline.failOnMissing>
+		<bnd.baseline.fullReport>false</bnd.baseline.fullReport>
+		<bnd.baseline.base.version>(,${project.version})</bnd.baseline.base.version>
 	</properties>
 
 	<scm>
@@ -66,6 +70,18 @@
 		</snapshotRepository>
 	</distributionManagement>
 
+	<repositories>
+		<repository>
+			<id>zsmartsystems</id>
+			<url>https://dl.bintray.com/zsmartsystems/com.zsmartsystems/</url>
+			<releases/>
+		</repository>
+		<repository>
+			<id>jcenter</id>
+			<url>https://jcenter.bintray.com/</url>
+			<releases/>
+		</repository>
+	</repositories>
 	<modules>
 		<module>com.zsmartsystems.zigbee</module>
 		<module>com.zsmartsystems.zigbee.autocode</module>
@@ -215,11 +231,10 @@
 					</archive>
 				</configuration>
 			</plugin>
-
 			<plugin>
 				<groupId>org.apache.felix</groupId>
 				<artifactId>maven-bundle-plugin</artifactId>
-				<version>3.5.0</version>
+				<version>4.2.1</version>
 				<executions>
 					<execution>
 						<id>bundle-manifest</id>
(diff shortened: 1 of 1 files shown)


cdjackson:
I'm a bit unclear why this is needed and how it decides if the framework is "breaking semantic versioning rules".

Why do you think this is needed - I don't want to add things to the build if unnecessarily, but if it's required, then we can look at it (I've not come across the need for this in many years of working on OSGi projects).

sesuncedu:
I will add some reports of issues for a few different releases.

Semantic versioning rules are defined in terms of release to release binary compatibility. Packages with MAJOR changes are not substituted for older versions. MINOR versions can be substituted. MICRO versions do not have api changes. Compatibility is defined in the JLS.

cdjackson:
Thanks. My question wasn't what symantic versioning is - that I know. My question was about this maven goal. What does it do to check this, and why is it really required here?

I don't see the need for this, and I don't want to add something that is going to restrict the way the project is operating.

sesuncedu:
The baseline goal extracts the binary API by processing the  generated bundle file  and the bundle file for specified baseline version, and compares the two, classifying the changes based on the backwards/forwards compatibility standards. 

OSGI packages and bundle versions **MUST** follow semantic versioning. This is a core part of the OSGI specification, and OSGI frameworks rely on this information when wiring packages  

I have run in to this issue with ZSS / OH with e.g. console bundles generating class not found exceptions. 

Nothing in this process will restrict the way the project operates; it just requires that the version numbers be incremented correctly.

cdjackson:
It is still unclear to me what this does. I don't want this to force me to increment the major version at this stage of the development.

What does it do to check this, and why is it really required here?

sesuncedu:
It parses the class files. It looks at what has changed in the api between the two versions.

 If an API change has been made that is not binary backwards compatible, it fails the build if the appropriate major version has not increased.

If an change  API change but it is binary backwards compatible, is fails the build if the appropriate minor version has not been increased. 
### Example - com.zsmartsystems.zigbee, version 1.2.1 vs 1.2.0 
Taking  the bundle ```com.zsmartsystems.zigbee``` as an example, and comparing bundle version ```1.2.1``` with ```1.2.0``` using the bnd command line tool (which uses the same code for baseline as the maven plugins). 

Running  baseline shows the following errors/changes
```
===============================================================
* com.zsmartsystems.zigbee 1.2.1-1.2.0 suggests 2.0.0
===============================================================
  Package                                            Delta      New        Old        Suggest    If Prov.  
* com.zsmartsystems.zigbee                           MAJOR      1.2.1      1.2.0      2.0.0      -         
* com.zsmartsystems.zigbee.app.iasclient             MAJOR      1.2.1      1.2.0      2.0.0      -         
* com.zsmartsystems.zigbee.app.otaserver             MAJOR      1.2.1      1.2.0      2.0.0      -         
* com.zsmartsystems.zigbee.zcl                       MINOR      1.2.1      1.2.0      1.3.0      -         
* com.zsmartsystems.zigbee.zcl.clusters              MINOR      1.2.1      1.2.0      1.3.0      -         
  com.zsmartsystems.zigbee.zcl.clusters.greenpower   ADDED      1.2.1      -          ok         -         
* com.zsmartsystems.zigbee.zcl.protocol              MINOR      1.2.1      1.2.0      1.3.0      -         
```

We see that there are three packages that have at least one change that is binary backwards incompatible. 

Let's take a look at the package ```com.zsmartsystems.zigbee```
There are a couple of classes with breaking changes.  
For example, ZigBeeNetworkManager has an incompatible API change - a public method is removed.
 ```
 MAJOR      CLASS      com.zsmartsystems.zigbee.ZigBeeNetworkManager
  REMOVED    METHOD     addNode(com.zsmartsystems.zigbee.ZigBeeNode)
```
Another breaking change is in ```ZigBeeNode```.  The return type of a method is changed from void to boolean.  
This change is *source-code* backwards compatible; however in java, the result type is part of the method name, so code compiled against an older version of the library will fail at runtime.
```
 MAJOR      CLASS      com.zsmartsystems.zigbee.ZigBeeNode
  MAJOR      METHOD     setNetworkAddress(java.lang.Integer)
   ADDED      RETURN     boolean
   REMOVED    RETURN     void
```

The package ```com.zsmartsystems.zigbee.zcl``` has only a minor change. The class ```ZclAttribute``` has a method added.  Existing compiled code will work fine;  new code that uses this package requires this version or later. 
```
MINOR      PACKAGE    com.zsmartsystems.zigbee.zcl
 MINOR      CLASS      com.zsmartsystems.zigbee.zcl.ZclAttribute
  ADDED      METHOD     writeValue(java.lang.Object)
```

cdjackson:
Ok, thanks for the explanation. Unfortunately at this time I can't really accept this while in this stage of development as I don't want to be updating the version.

sesuncedu:
Bundles that import packages usually specify an import range that starts from the minor version compiled against up to, but not including the next major package version.  

For example, the bundle ```org.openhab.binding.zigbee``` imports the package ```com.zsmartsystems.zigbee``` with version range ```[1.2,2)```.

cdjackson:
Yes, I know, but how does that change anything?

cdjackson:
@sesuncedu is there anything further to discuss with this or shall we close it?

cdjackson:
Closing this - if there's anything further to discuss, please provide information requested.

Issue #791, opened 2019-10-10
Ember: Race conditions assigning sequence numbers in  EzspFrameRequest


sesuncedu:
EZSP frame sequence numbers are integers in the range [1,254].  

EzspFrameRequest  stores the value of the next sequence number to assign  in an ```AtomicLong```.

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/d7398dd5edcfb5f78a3e367a82146ea7d73d95a2/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ezsp/EzspFrameRequest.java#L43

Sequence numbers are assigned in the constructor

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/d7398dd5edcfb5f78a3e367a82146ea7d73d95a2/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ezsp/EzspFrameRequest.java#L48-L53

This code is subject can return invalid values due to the non-atomic manipulation of the sequence when the value overflows.  For example, if a thread is assigned sequence number 254, but other threads increment the value before the first thread has had a chance to reset it to 1, the those threads can  be assigned sequence numbers of 255 or higher.  

The fix is to get and update the sequence in a single atomic operation. 
e.g.
```java
    private final static AtomicInteger sequence = new AtomicInteger(1);

    /**
     * Constructor used to create an outgoing frame
     */
    protected EzspFrameRequest() {
        sequenceNumber = sequence.getAndUpdate(n -> n <254 ? n+1 : 1);
    }
```

I need to stash some stuff so I can create a  PR with just this change.

cdjackson:
Thanks. As commented on the PR, I’d suggest to just synchronise the existing code. We can probably change to using an AtomicInteger - I think in the past I needed to keep track of the total number of frames sent which is why it is done the way it is, but that’s not used now, but we can’t use streaming.

sesuncedu:
Using getAndUpdate is faster and less typing, so what's not to like :) 

[Being able to use streaming again would be nice  if setSourceRoute is used.]

cdjackson:
Unfortunately streaming is not possible due to Android limitations.

sesuncedu:
I'm confused about android. Is this an issue with android usbserial drivers? 
[...]

PR #792, opened 2019-10-10
Avoid a race condition when assigning sequence numbers in the constructor of EzspFrameRequest.

sesuncedu:
See: #791 

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ezsp/EzspFrameRequest.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ezsp/EzspFrameRequest.java
@@ -7,6 +7,7 @@
  */
 package com.zsmartsystems.zigbee.dongle.ember.ezsp;
 
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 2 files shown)


sesuncedu:
Sigh: tests aren't running in a fresh VM.

com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ezsp/EzspFrameRequest.java
```diff
@@ -40,16 +41,13 @@
...
 
     /**
      * Constructor used to create an outgoing frame
      */
     protected EzspFrameRequest() {
-        sequenceNumber = (int) sequence.getAndIncrement();
-        if (sequenceNumber == 254) {
-            sequence.set(1);
-        }
+        sequenceNumber = sequence.getAndUpdate(n -> n <254 ? n+1 : 1);
```
cdjackson: Please use the old notation and synchronise the method for backward compatibility.

cdjackson:
Did you mean to close this, or do you plan to provide this in a separate PR?

sesuncedu:
Separate PR with the stupidity squashed.

Also, I am now not 100% sure that the frame request  is the right place to be assigning sequence numbers in the first place (if the requirement is that they be presented to the ncp in sequential order, then they should be assigned when they are transmitted, since the order in which they are presented may not be the same as the one in which they were created.

cdjackson:
Fair point - feel free to change this to go into the ASH/SPI handlers?

PR #793, opened 2019-10-11
Avoid race condition when assigning sequence numbers in EzspFrameRequest - #791

sesuncedu:
Use ```AtomicInteger::getAndUpdate``` to handle sequence number wrap.


cdjackson:
Looks like the tests failed. Also, please run the formatter over the code...

I guess your point in the other thread still remains - should we just migrate the setting of the sequence number to the ASH and SPI handlers so that they remain in order - I think that is likely the best approach if we think this is required (although the EZSP doc doesn't seem to explicitly state this, but I think it is probably the intent and in any case it is probably safer).

sesuncedu:
[Sigh - fire alarm and cable modem fun.]

Section 3 of the Ezsp reference 
>The first byte of all EZSP frames is a sequence number. The host should increment the sequence number each time a command is sent to the NCP.

Also, I can't see any restrictions on the value of the sequence number? In which case can just mask a getAndIncrement with 0xff.

cdjackson:
> Section 3 of the Ezsp reference

This is what I was referring to - in my interpretation it doesn't explicitly state that they must be in order. This might just be my interpretation though and probably it is the intent that they should be in order.

> case can just mask a getAndIncrement with 0xff.

It might be ok - I wrote this a long time ago now and I don't recall why I skipped 255 and 0.

com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java
```diff
@@ -429,6 +432,8 @@ private synchronized boolean sendNextFrame() {
             // Nothing to send
             return false;
         }
+        int sequenceNumber = nextEzspSequenceNumber.getAndIncrement() & 0xff;
```
cdjackson: Since ```sequenceNumber``` doesn't appear to be used at all (other than in the next line) it might be better not to create this variable and directly set the sequence number?

```
nextFrame.setSequenceNumber(nextEzspSequenceNumber.getAndIncrement() & 0xff);
```

com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandler.java
```diff
@@ -522,7 +524,8 @@ private synchronized boolean sendNextFrame() {
                 return false;
             }
         }
-
+        int sequenceNumber = nextEzspSequenceNumber.getAndIncrement() & 0xff;
```
cdjackson: Since ```sequenceNumber``` doesn't appear to be used at all (other than in the next line) it might be better not to create this variable and directly set the sequence number?

```
nextFrame.setSequenceNumber(nextEzspSequenceNumber.getAndIncrement() & 0xff);
```

com.zsmartsystems.zigbee.dongle.ember/src/test/java/com/zsmartsystems/zigbee/dongle/ember/internal/spi/SpiFrameHandlerTest.java
```diff
@@ -277,25 +278,36 @@ public void testConnect() {
...
 
     @Test
-    public void testSendEzspFrame() {
+    public void testSendEzspFrame() throws NoSuchFieldException, IllegalAccessException {
         portOutData = new ArrayList<>();
         SpiFrameHandler handler = new SpiFrameHandler(Mockito.mock(EzspFrameHandler.class));
         handler.start(new TestPort(Mockito.mock(InputStream.class), Mockito.mock(OutputStream.class)));
 
+        final int sequenceNumber = 15;
+        setNextSequenceNumber(handler, sequenceNumber);
```
cdjackson: You might want to use the ```TestUtilities``` class - it has a helper to set private fields. Since I fear in future this might all need to change, using the single method is probably useful.

cdjackson:
Thanks @sesuncedu - looks good - just one minor comment.

cdjackson:
Why are you closing this? I think it's still needed isn't it?

cdjackson:
@sesuncedu can you please advise why this is closed? Do you plan to open a new PR, or what is the status?

PR #802, opened 2019-10-14
Improve handling of ezsp transactions.

sesuncedu:
Don't process messages that are consumed by a transaction to non-specific callback handler. 
Make responses are returned as results of futures.
Make transactions generic, with terminal response as parameter.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java
@@ -449,7 +449,7 @@ public class AshFrameHandler implements EzspProtocolHandler {
             case DATA:
                 statsTxData++;
                 // Set the frame number
-                ((AshFrameData) ashFrame).setFrmNum(frmNum);
+                ashFrame.setFrmNum(frmNum);
                 frmNum = (frmNum + 1) & 0x07;
 
                 // DATA frames need to go into a sent queue so we can retry if needed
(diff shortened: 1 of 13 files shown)


cdjackson:
Thanks for this @sesuncedu. Generally I'm happy with the generics and general improvements (thanks :) ).

Can you describe why you want to make these changes? (I'm not talking about the generics etc). I actually thought that the system didn't pass up messages that were fullfilled by a transaction now, so maybe this is a bug that you're fixing (?).

Why change the transaction matching? I've not looked at this too much yet but it looks quite different (it looks to be not using sequence numbers for matching the EZSP responses, although maybe I need to look over this more closely) and it would be really good to describe the change in the issue so we have some (simple) record what changed and why in future.

Just as a general comment - if possible (and I appreciate it doesn't always work this way) can you break each change into separate PRs so it's easier to review. This PR has loads of changes which are mostly the generics/futures and it makes it harder to find the other changes.

com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java
```diff
@@ -202,8 +202,10 @@ public void run() {
                                         if (response == null) {
                                             logger.debug("ASH: No frame handler created for {}", packet);
                                         } else {
-                                            notifyTransactionComplete(response);
-                                            handleIncomingFrame(response);
+                                            if (!notifyTransactionComplete(response)) {
```
sesuncedu: Before, the result notifyTransactionComplete was ignored, and the response message unconditionally on to handleIncomingFrame, even if it was a (completing) match.

sesuncedu:
Why separate matching from completion:

Previously, the message handling code returned true if a message was matched and it completed the transaction, and false otherwise (either the message was not  a match, or it was a match but did not complete the transaction). 

The use of a single value to cover both cases meant that there was no way to signal only the fact that a message had been consumed.  Thus the  result of e.g. AshFrameHandler::notifyTransactionComplete was always ignored, and default processing was always run, even when a response was consumed.

cdjackson:
We need to check this - there was a reason that the response was ignored. In the early days, it wasn't ignored, but more recently we needed to pass up all frames even if they were consumed by an EZSP transaction and I don't remember why this is, but before making this change we ought to check if possible.

cdjackson:
Why is this closed? Is it not needed? I would really prefer to try and stick with the original PRs than closing and opening new ones to avoid additional work with the reviewing - if you must close them, please provide a link to any new PR.

PR #804, opened 2019-10-14
Combined extended timeout and EZSP Transaction branch

sesuncedu:
A combined PR that merges the EZSP transaction and extended timeout changes.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/EmberCbkeProvider.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/EmberCbkeProvider.java
@@ -290,9 +290,9 @@ public class EmberCbkeProvider implements ZigBeeCbkeProvider {
         request.setMyKey(myKey);
         request.setMyCert(myCert);
         request.setCaCert(caCert);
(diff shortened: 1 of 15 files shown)


sesuncedu:
Note: currently unhandled ezsp messages are logged at the WARN level.  It seems like this is the right level, since it doesn't seem to quite rise to the level of error, but doesn't seem to be just INFO.

The main ones I'm seeing are: 
- Network state responses from the keepalive poll, 
- IncomingRouteRecordHandler messages from having concentrator turned on but source routing not active 

I will add a commit  for the former.

cdjackson:
As I mentioned earlier, please avoid creating large PRs with lots of features. I really MUCH prefer smaller PRs per additional feature - I know that can sometimes be more painful to test, but a) it's easier to review, and b) it's easier to see the progression of the product, and c) it's easier to cherry pick if users don't want some features due to breaking changes in their product.

Please can you split these? If not, then at least we should close this and stick with the current PRs and in future try and keep the PRs to a single issue.

Thanks.

sesuncedu:
There are two pre-existing split ones, and this one which fixes a merge issue between the two. 
[doing stuff that is all hitting around the same areas, and which has dependencies, so I do this to myself]

One of the unmerged ones has a spurious code-coverage fail. It's fixed in this PR by doing absolutely nothing.

cdjackson:
Ok, so shall we close this then and work through the comments in the other PRs?

PR #828, opened 2019-10-24
Return responses for async requests on ezsp frame handlers

sesuncedu:
Asynchronous requests to EZSP Frame handlers always returned null from the call used to produce the value of their callables. 

(this change has been separated from a previous transaction related pull request).


sesuncedu:
Note: Only ASH variant has been tested in production.

Issue #837, opened 2019-10-26
Responses to outgoing messages addressed to clusters for which explicit support has not been registered appear to be rejected


sesuncedu:
The changes in #826 appear to be causing the framework to reject incoming messages that are from clusters that the framework has sent messages to. 

I have created a PR #836 with a little  debugging code to illustrate. 

Maybe #826 should be backed out for now?

#### Read Attributes Command for On/Off
```
17:55:52.572 [DEBUG] [ing.zigbee.handler.ZigBeeThingHandler] - 0022A30000252FBF: Polling zigbee:device:536f0a01:0022a30000252fbf:0022A30000252FBF_1_switch
17:55:52.575 [DEBUG] [rtsystems.zigbee.ZigBeeNetworkManager] - TX CMD: ReadAttributesCommand [On/Off: 0000/0 -> 4092/1, cluster=0006, TID=F4, identifiers=[0]]
17:55:52.577 [DEBUG] [rtsystems.zigbee.ZigBeeNetworkManager] - TX ZCL: ZclHeader [frameType=ENTIRE_PROFILE_COMMAND, manufacturerSpecific=false, direction=CLIENT_TO_SERVER, disableDefaultResponse=false, manufacturerCode=0, sequenceNumber=244, commandId=0]
17:55:52.579 [DEBUG] [rtsystems.zigbee.ZigBeeNetworkManager] - TX APS: ZigBeeApsFrame [sourceAddress=0/1, destinationAddress=16530/1, profile=0104, cluster=0006, addressMode=DEVICE, radius=31, apsSecurity=false, apsCounter=24, payload=00 F4 00 00 00]
17:55:52.582 [DEBUG] [le.ember.internal.ash.AshFrameHandler] - TX EZSP: EzspSendUnicastRequest [type=EMBER_OUTGOING_DIRECT, indexOrDestination=4092, apsFrame=EmberApsFrame [profileId=0104, clusterId=0006, sourceEndpoint=1, destinationEndpoint=1, options=[EMBER_APS_OPTION_ENABLE_ROUTE_DISCOVERY, EMBER_APS_OPTION_ENABLE_ADDRESS_DISCOVERY, EMBER_APS_OPTION_RETRY], groupId=0, sequence=24], messageTag=F4, messageContents=00 F4 00 00 00]
17:55:52.585 [DEBUG] [le.ember.internal.ash.AshFrameHandler] - ASH: TX EZSP queue size: 1
[...]

PR #838, opened 2019-10-26
Partially revert #826 by not rejecting messages to unsupported client clusters. ISSUE #837

sesuncedu:
These messages include responses to commands sent to servers.

This change does not change the results of any tests, but prevents desired responses from being dropped, such as read attribute responses and attribute reporting triggered by   device discovery or channel drivers when deployed in openhab.

I have not checked whether there are issues relating to rejected server commands remaining.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
@@ -968,11 +968,6 @@ public class ZigBeeNetworkManager implements ZigBeeNetwork, ZigBeeTransportRecei
 
         ZclCommand command;
         if (zclHeader.getDirection() == ZclCommandDirection.SERVER_TO_CLIENT) {
-            if (clusterMatcher != null && !clusterMatcher.isClientSupported(apsFrame.getCluster())) {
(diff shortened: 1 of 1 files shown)


cdjackson:
Sorry - as I've already said in the other PR you created, this cannot be reverted. You need to change your application to correctly reflect the clusters that your application supports. Doing this will cause test failures so I can't revert this.

sesuncedu:
There does not appear to be a server related issue (there was some odd behavior from a light switch that has gone away)

sesuncedu:
No test cases fail when this is applied,

sesuncedu:
Broken code includes discovery and any openhab bindings that use commands that need responses,

sesuncedu:
Any application that reads attributes without registering a client cluster.

cdjackson:
If you are trying to use openHAB with the latest framework, yes, this will not work unless you have modified it to work with the latest framework. Did you modify your system to add support for all clusters that you are planning to use in your application? I would guess not?

The framework must not respond to any clusters or endpoints that it does not state support for in the descriptor. If it does, it WILL fail certification tests. You state that it doesn't fail these tests - what test harness are you using to verify this?

cdjackson:
@sesuncedu can we close this - is this just your misunderstanding or is there still an issue?

cdjackson:
No response from @sesuncedu after 2 months so closing this as it seems to be a misunderstanding about how the framework works.

Issue #875, opened 2019-11-22
Queue in ZigbeeTransactionManager is blocked by old transaction


triller-telekom:
- Pair a device with for example openHAB.
- Remove the device via openHAB (we used a battery powered one and removed the battery before deleting it to make the leave request fail)
- Try to pair the device again.

The problem we found is in the `ZigbeeTransactionManager` whose `nodeUpdated()` get called so he can clean up all transactions corresponding to the node. We saw that the queue for the removed device contains a transaction with a destination address (network address) that is the OLD address of the device we are trying to pair again.

All new transactions, like discovery of descriptors or cluster bindings are queued AFTER this old transaction. The policy for sending data has a `maxOutstandingTranscations=1` and thus the queue is effectively blocked.

I made a proof of concept:

In the `ZigbeeTransactionManager#nodeUpdated()` I call a new method on the `ZigBeeTransactionQueue` and I named it 

```
void removeTransactionsNotBelongingToNode(ZigBeeNode node) {
        ZigBeeTransaction zt = null;
        boolean allCleared = false;
        while ((zt = queue.peek()) != null && !allCleared) {
            ZigBeeAddress destinationAddress = zt.getDestinationAddress();
            if (destinationAddress != null) {
                if (destinationAddress.getAddress() != node.getNetworkAddress()) {
                    // we found an old transaction which should not be send to this updated node anymore -> remove it
                    // from the queue
                    logger.debug("Stefan: Removing old transaction from queue: {}", zt);
                    queue.poll();
                } else {
                    // found the point in the queue were only new transactions are
                    allCleared = true;
                }
            } else {
                logger.debug("No destination address set in transaction={}", zt);
            }
        }
    }
```

This method removes all transactions from the queue which are not targeted towards the networkaddress of the current (updated) node.

After cleaning the queue, I am able to pair the device again.

Please see the log attached.

[log_clean.txt](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/files/3879776/log_clean.txt)

cdjackson:
Thanks Stefan,

I guess there are two possible solutions to this - one is as you've suggested, to clear the queue, but I'm not sure this is the right solution. The other solution is to change the destination network address to the updated value.

[...]

PR #897, opened 2019-12-09
Add IEEE address of the device to ZigBeeTransaction and ZigBeeTransactionQueue

triller-telekom:
Contributes to #875

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionManager.java
@@ -168,14 +168,14 @@ public class ZigBeeTransactionManager implements ZigBeeNetworkNodeListener {
         defaultProfile = new ZigBeeTransactionProfile(NODE_RETRIES, NODE_TRANSACTIONS, NODE_DELAY);
         defaultSleepyProfile = new ZigBeeTransactionProfile(SLEEPY_RETRIES, SLEEPY_TRANSACTIONS, SLEEPY_DELAY);
 
-        defaultQueue = new ZigBeeTransactionQueue("Default", null);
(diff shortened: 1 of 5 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionQueue.java
```diff
@@ -161,6 +168,15 @@ public boolean isSleepy() {
         return sleepy;
     }
 
+    /**
+     * Gets the {@link IeeeAddress} of the device for which this queue has transactions
+     *
+     * @return the IEEE address
```
cdjackson: It might pay to add that this may be null if the queue is a generic queue not linked to a device?

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionManager.java
```diff
@@ -168,14 +168,14 @@ public ZigBeeTransactionManager(ZigBeeNetworkManager manager) {
         defaultProfile = new ZigBeeTransactionProfile(NODE_RETRIES, NODE_TRANSACTIONS, NODE_DELAY);
         defaultSleepyProfile = new ZigBeeTransactionProfile(SLEEPY_RETRIES, SLEEPY_TRANSACTIONS, SLEEPY_DELAY);
 
-        defaultQueue = new ZigBeeTransactionQueue("Default");
+        defaultQueue = new ZigBeeTransactionQueue("Default", null);
```
cdjackson: I think it's cleaner to have a separate constructor rather than passing in null for the address?

triller-telekom:
Thanks @cdjackson for your feedback. I have addressed your comments, you may have another look.

cdjackson:
Thanks @triller-telekom

PR #898, opened 2019-12-11
Rewrite transaction destinations if node rejoins with new network address

### LINKED ISSUE #875: Queue in ZigbeeTransactionManager is blocked by old transaction

triller-telekom:
- to be merged AFTER #897 (compilation depends on it)

Fixes #875

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionManager.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionManager.java
@@ -12,6 +12,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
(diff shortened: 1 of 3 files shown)


triller-telekom:
Since #897 has been merged, I have rebased this PR.

@cdjackson Can you please have a look at this solution, whether you think it is appropriate to fix #875 ?

cdjackson:
Thanks Stefan - looks good to me.

PR #933, opened 2019-12-27
Ota on cc2531

wsowa:
I have done some work to make OTA Server work reliably on my setup which is CC2531 dongle with a custom ZR and ZED based on NXP's ZigBee stack for JN5169 (which OTA client behavior I have verified against the ZigBee spec and it looks fine).

I have split it into a number of commits each bringing some value independently yet all of them contribute to my goal.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.conbee/src/main/java/com/zsmartsystems/zigbee/dongle/conbee/ZigBeeDongleConBee.java
+++ b/com.zsmartsystems.zigbee.dongle.conbee/src/main/java/com/zsmartsystems/zigbee/dongle/conbee/ZigBeeDongleConBee.java
@@ -300,6 +300,7 @@ public class ZigBeeDongleConBee implements ZigBeeTransportTransmit {
 
         request.setRequestId(apsFrame.getApsCounter());
         request.setClusterId(apsFrame.getCluster());
+        request.setTxOptions(apsFrame.getAckRequest() ? 0x40 : 0);
         switch (apsFrame.getAddressMode()) {
             case DEVICE:
                 request.setDestinationAddress(
@@ -320,7 +321,6 @@ public class ZigBeeDongleConBee implements ZigBeeTransportTransmit {
         request.setProfileId(apsFrame.getProfile());
         request.setRadius(apsFrame.getRadius());
         request.setSourceEndpoint(apsFrame.getSourceEndpoint());
-        // request.setTxOptions(txOptions);
 
         request.setAdsuData(apsFrame.getPayload());
 
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java
@@ -684,7 +684,9 @@ public class ZigBeeDongleEzsp implements ZigBeeTransportTransmit, ZigBeeTranspor
         emberApsFrame.setSourceEndpoint(apsFrame.getSourceEndpoint());
         emberApsFrame.setDestinationEndpoint(apsFrame.getDestinationEndpoint());
         emberApsFrame.setSequence(apsFrame.getApsCounter());
-        emberApsFrame.addOptions(EmberApsOption.EMBER_APS_OPTION_RETRY);
+        if (apsFrame.getAckRequest()) {
+            emberApsFrame.addOptions(EmberApsOption.EMBER_APS_OPTION_RETRY);
+        }
         emberApsFrame.addOptions(EmberApsOption.EMBER_APS_OPTION_ENABLE_ROUTE_DISCOVERY);
         emberApsFrame.addOptions(EmberApsOption.EMBER_APS_OPTION_ENABLE_ADDRESS_DISCOVERY);
 
--- a/com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/ZigBeeDongleTelegesis.java
+++ b/com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/ZigBeeDongleTelegesis.java
@@ -342,7 +342,7 @@ public class ZigBeeDongleTelegesis
     private final int defaultS10 = 0x56A9;
 
     /**
-     * Map used to correlate the Telegesis sequence IDs with the APS counter we use to correlate messages in the stack
+     * Map used to correlate the Telegesis sequence IDs with the Transaction ID we use to correlate messages in the stack
      */
     private Map<Integer, Integer> messageIdMap = new ConcurrentHashMap<>();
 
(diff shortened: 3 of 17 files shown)


cdjackson:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransaction.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/zsmartsystems/com-zsmartsystems-zigbee/pullRequest?prid=4736197&bid=15810449)

com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/ZigBeeDongleTelegesis.java
```diff
@@ -342,7 +342,7 @@
     private final int defaultS10 = 0x56A9;
 
     /**
-     * Map used to correlate the Telegesis sequence IDs with the APS counter we use to correlate messages in the stack
+     * Map used to correlate the Telegesis sequence IDs with the Transaction ID we use to correlate messages in the stack
```
wsowa: We actually use transaction ID as a key in that map and in equivalents in the other dongles code. This comment mislead me a bit so correcting

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/aps/ZigBeeApsFrame.java
```diff
@@ -78,6 +78,12 @@
      */
     private boolean securityEnable;
 
+    /**
+     * The AckRequest parameter specifies whether the current transmission requires an acknowledgement frame to be sent
+     * be the recipient on receipt of the frame.
+     */
+    private boolean ackRequest;
```
wsowa: One note: It seems that Telegesis API doesn't support disabling ACK request in the APS frame. Other dongles looks like they do but I only had a chance to test on CC2531

com.zsmartsystems.zigbee.dongle.conbee/src/main/java/com/zsmartsystems/zigbee/dongle/conbee/ZigBeeDongleConBee.java
```diff
@@ -300,6 +300,7 @@ public void sendCommand(final int msgTag, final ZigBeeApsFrame apsFrame) {
 
         request.setRequestId(apsFrame.getApsCounter());
         request.setClusterId(apsFrame.getCluster());
+        request.setTxOptions(apsFrame.getAckRequest() ? 0x40 : 0);
```
wsowa: Please, can you test that the frame is indeed transmitted with APS Ack Request = FALSE as I don't have a dongle to test with.

com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java
```diff
@@ -684,7 +684,9 @@ public void sendCommand(final int msgTag, final ZigBeeApsFrame apsFrame) {
         emberApsFrame.setSourceEndpoint(apsFrame.getSourceEndpoint());
         emberApsFrame.setDestinationEndpoint(apsFrame.getDestinationEndpoint());
         emberApsFrame.setSequence(apsFrame.getApsCounter());
-        emberApsFrame.addOptions(EmberApsOption.EMBER_APS_OPTION_RETRY);
+        if (apsFrame.getAckRequest()) {
+            emberApsFrame.addOptions(EmberApsOption.EMBER_APS_OPTION_RETRY);
```
wsowa: Please, can you test that the frame is indeed transmitted with APS Ack Request = FALSE as I don't have a dongle to test with.

com.zsmartsystems.zigbee.dongle.xbee/src/main/java/com/zsmartsystems/zigbee/dongle/xbee/ZigBeeDongleXBee.java
```diff
@@ -337,6 +337,10 @@ public void sendCommand(final int msgTag, final ZigBeeApsFrame apsFrame) {
             command.addOptions(TransmitOptions.ENABLE_APS_ENCRYPTION);
         }
 
+        if (!apsFrame.getAckRequest()) {
+            command.addOptions(TransmitOptions.DISABLE_RETRIES);
```
wsowa: Please, can you test that the frame is indeed transmitted with APS Ack Request = FALSE as I don't have a dongle to test with.

com.zsmartsystems.zigbee.dongle.conbee/src/main/java/com/zsmartsystems/zigbee/dongle/conbee/ZigBeeDongleConBee.java
cdjackson: I don't know anyone using this coordinator

cdjackson:
Hi @wsowa. Thanks for this. I think we really need to split this into separate PRs - not just separate commits. This helps visibility, but also I need to be careful how things here are merged as I am currently performing certification, and features like fragmentation have already been verified by the ZigBee certification tests house. (I've been through 3 days of pre-certification testing, and now just need to do the formal tests for a Smart Energy system).

Please can you split these into logical PRs? If you can fix simple bugs such as incorrect transaction id and the fragmentation issue in separate PRs, that would definitely be good.

Why are you disabling APS ACKs here? Again, I think this should be a separate PR.

wsowa:
Hi, no problem, I'll get separate PRs for this. I assume you want them originating from the same base?

Fragmentation issue I'm fixing here will manifest itself in quite specific scenario with incomplete transactions. Scenario is more or less you send multiple frames with payload length at the boundary of fragmentation (more than 256 so you have stale incomplete frames in fragmentTxQueue) over a short period of time (a few seconds). After that you send a short frame for which msgTag you'll get an incomplete frame in the fragmentTxQueue. Just looking at the condition that determines if the payload fits the frame it is clear the condition should be <= rather than <
`apsFrame.getPayload().length < fragmentationLength`
I'll try to add a test case to demonstrate what the problem is in the PR long with that fix. As I said it is quite specific scenario so perhaps it wasn't verified in the test house.

I disable APS ACKs only for OTA Image Block Response Command when it is a response to a Image Page Req (For Image Page Request the server sends multiple blocks rather then one as it is a case for Image Block Request). The OTA spec says.
> In response to the Image Page Request, the server shall send Image Block Response commands with no APS retry to disable APS acknowledgement. The intention is to minimize the number of packets sent by the client in order to optimize the energy saving. APS acknowledgement is still used for Image Block Response sent in response to Image Block Request command.

You actually had a comment `// TODO: Ideally we should disable APS retry for page requests` in ZclOtaUpgradeServer.java which I removed with my commit :)

Expect the PRs over next couple of days.

Best regards,
Witek

cdjackson:
Thanks - that all sounds good. Yes, all PRs from the same base would be appreciated, and any tests would be perfect :)

Cheers
Chris

wsowa:
Hi. All parts of this PR expect commit e931359 (OTA: Send Image Blocks without APS ACKs requests when replying to Page Request) are passed in separate PR. The one missing is dependent on the changes provided in the PRs #936 and #940 so I'll wait for them to get merged before passing in that one. Unless you suggest different course of action?

cdjackson:
Thanks @wsowa that sounds fine and I will try and get all these processed in the next 2 days.

cdjackson:
@wsowa is everything here now covered?  I think we still need to address APS ACKs, but I don't think this PR addressed that anyway?

If so, can we now close this?

wsowa:
APS ACKs are in this PR but they are depent of OTA server fixes in PR #964. I can see they got merged yesterday. Let's close this PR and I'll issue a new one for APS ACKs within next few days. Thanks.

cdjackson:
Great - thanks.  I think once that is done I will release 1.3.0 and update the OH binding.

PR #936, opened 2019-12-27
OTA: Send OTA Response commands with Disable Default Response = True

wsowa:
The OTA specification requires that all Request Commands sent by the client have Disable Default Response = False i.e. require ZCL respone with the same transaction seq number as the request. Moreover the Response commands should have Disable Default Response = True. We were incorrectly replying with different seq number and with Disable Default Response = False. That wouldn't be that bad if the client would reply to the response with the ZCL default response. Yet the client I was working with was quite fussy about it and decided to not reply (we were first not compliant with the standard so I don't blame it) leaving us with incomplete transactions which quickly got stuck in the queue until timeout which in turn caused the OTA process to freeze and timeout finally.
With this change we become compliant with the standard by sending responses with the seq id matching the one in the request and with disabled default response and also we complete the transaction on RX_ACK as we don't expect the response.

### DIFF
--- a/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeZclClusterGenerator.java
+++ b/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeZclClusterGenerator.java
@@ -21,6 +21,8 @@ import com.zsmartsystems.zigbee.autocode.xml.ZigBeeXmlCluster;
 import com.zsmartsystems.zigbee.autocode.xml.ZigBeeXmlCommand;
 import com.zsmartsystems.zigbee.autocode.xml.ZigBeeXmlField;
 
+import static java.util.stream.Collectors.joining;
+
 /**
  *
  * @author Chris Jackson (zsmartsystems.com)
@@ -390,27 +392,39 @@ public class ZigBeeZclClusterGenerator extends ZigBeeBaseClassGenerator {
             out.println("     *");
 
             final LinkedList<ZigBeeXmlField> fields = new LinkedList<ZigBeeXmlField>(command.fields);
+
+            if (command.responseTo.isPresent()) {
+                out.println("     * @param request {@link " + stringToUpperCamelCase(command.responseTo.get())
+                        + " A request that this command replies to");
+            }
             for (final ZigBeeXmlField field : fields) {
                 out.println("     * @param " + stringToLowerCamelCase(field.name) + " {@link " + getDataTypeClass(field)
                         + "} " + field.name);
             }
 
-            out.println("     * @return the {@link Future<CommandResult>} command result future");
-            out.println("     */");
-            out.print("    public Future<CommandResult> " + stringToLowerCamelCase(command.name) + "(");
-
-            boolean first = true;
-            for (final ZigBeeXmlField field : fields) {
-                if (first == false) {
-                    out.print(", ");
-                }
-                out.print(getDataTypeClass(field) + " " + stringToLowerCamelCase(field.name));
-                first = false;
+            if (!command.responseTo.isPresent()) {
+                out.println("     * @return the {@link Future<CommandResult>} command result future");
             }
+            out.println("     */");
+            if (command.responseTo.isPresent()) {
+                out.print("    public void " + stringToLowerCamelCase(command.name) + "(");
+            } else {
+                out.print("    public Future<CommandResult> " + stringToLowerCamelCase(command.name) + "(");
+            }
+
+            String requestArg = command.responseTo.map(responseTo -> stringToUpperCamelCase(responseTo) + " request, ").orElse("");
+            String fieldsArgs = fields.stream()
+                    .map(field -> getDataTypeClass(field) + " " + stringToLowerCamelCase(field.name))
+                    .collect(joining(", "));
+            out.print(requestArg + fieldsArgs);
 
             out.println(") {");
             if (fields.size() == 0) {
-                out.println("        return send(new " + stringToUpperCamelCase(command.name) + "());");
+                if (command.responseTo.isPresent()) {
+                    out.println("        sendResponse(request, new " + stringToUpperCamelCase(command.name) + "());");
+                } else {
+                    out.println("        return send(new " + stringToUpperCamelCase(command.name) + "());");
+                }
             } else {
                 out.println("        " + stringToUpperCamelCase(command.name) + " command = new "
                         + stringToUpperCamelCase(command.name) + "();");
@@ -422,7 +436,11 @@ public class ZigBeeZclClusterGenerator extends ZigBeeBaseClassGenerator {
                             + stringToLowerCamelCase(field.name) + ");");
                 }
                 out.println();
-                out.println("        return send(command);");
+                if (command.responseTo.isPresent()) {
+                    out.println("        sendResponse(request, command);");
+                } else {
+                    out.println("        return send(command);");
+                }
             }
             out.println("    }");
         }
--- a/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/xml/ZigBeeXmlCommand.java
+++ b/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/xml/ZigBeeXmlCommand.java
@@ -8,6 +8,7 @@
 package com.zsmartsystems.zigbee.autocode.xml;
 
 import java.util.List;
+import java.util.Optional;
 
 /**
  *
@@ -17,6 +18,7 @@ import java.util.List;
 public class ZigBeeXmlCommand {
     public Integer code;
     public String source;
+    public Optional<String> responseTo;
     public String name;
     public List<ZigBeeXmlDescription> description;
     public List<ZigBeeXmlField> fields;
--- a/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/xml/ZigBeeXmlParser.java
+++ b/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/xml/ZigBeeXmlParser.java
@@ -11,6 +11,7 @@ import java.io.File;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.TreeMap;
 
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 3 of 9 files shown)


cdjackson:
@wsowa thanks for this. I've not really looked over it in any detail yet, but there is an issue that we need to look at before going too far...

Since you have modified the autogenerated code, and this will therefore be overwritten when the autocoder runs. I probably should look over this to see how this could be rewritten to avoid modification of the autocoded files. 

Either this could be through modification of the autocoder with some new construct if that makes sense, or we would need to find a way to code this outside of the ```ZclOtaUpgradeCluster``` class.

I'm happy to discuss if you have any thoughts or ideas...

wsowa:
Oh, I see. I'll take a look at it over next days and will let you know what my thoughts are. Can you just give me a one paragraph intro to autogenarator you're using? How do I run it, how do I specify which which class to generate, etc. I presume we're talking about content of `com.zsmartsystems.zigbee.autocode` maven module?

wsowa:
Ok. I ended up with a mix of auto generated code and special handling outside of `ZclOtaUpgradeCluster`. The command description in XML can now specify what request the response replies to and if such option is used then the `sendResponse` is generated instead of `send` in the cluster command method. For sending non-first Image Block Response for Image Page Request I've ended up with custom code in the OtaServer as this is quite OTA specific case.

wsowa:
Another simple alternative that crossed mind today would be to inherit from generated OTA cluster class and put the logic there without why change to the XMLs and the autogenerator code. That would raise the question why to have as autogenerator for OTA cluster in the first place so I like that solution less then the one I've committed above. 

Any thoughts @cdjackson?

On a side note, why do you auto generate the code for the clusters at all? It seems like you code the clusters in XML rather than in Java. What's the point?

cdjackson:
> On a side note, why do you auto generate the code for the clusters at all?

Do you mean "why not just process this at runtime instead of creating the Java"? If so, I agree this is an option, but for debugging it is a lot tidier to generate the code (IMHO).

I will take a look at this over the next couple of days - from your comment, it seems the change to the autocoder is ok, but I would like to take a better look.

Thanks.

cdjackson:
@wsowa just to let you know I haven't forgotten about this. I had a bit of a look at this yesterday as I'm creating a firmware provider for OH2. So, I understand the issue but want to have a look at how this is done. 

Generally I don't have any real problem with what you've done - it looks fine, but I just want to have a bit of a wider think about these interfaces and scalability. I will look at it this week though.

cdjackson:
@wsowa here's what I'm thinking - this may also be linked to your question "why do you auto generate the code for the clusters at all" which I may have misinterpreted previously...

I am thinking that we don't need your changes to the autocoder. I think you only made this change so that you can call ```sendResponse``` which is currently ```protected```. I was already thinking before this issue of making both the ```sendCommand``` and ```sendResponse``` method ```public``` to support sending manufacturer specific extensions. In this PR, you make the ```sendCommand``` public (it is at this point just called ```send``` but I want to rename it for consistency) so that you can use it in the ```sendImageBlock``` method. If we also make ```sendResponse``` public, then we don’t need to use the ```cluster``` methods at all (and I would probably then deprecate them).

On top of this, I am thinking about changing the autocoder to return the command for all the setter command calls (that currently return ```void```) so it can be chained and used in a ```Builder``` style as below -:

From -:

```
        cluster.imageBlockResponse(request, ZclStatus.SUCCESS, otaFile.getManufacturerCode(), otaFile.getImageType(),
                otaFile.getFileVersion(), fileOffset, imageData);
```

To -:

```
        cluster.sendResponse(request, new ImageBlockResponse().setStatus(ZclStatus.SUCCESS)
                .setManufacturerCode(otaFile.getManufacturerCode())
                .setOtaFile(otaFile.getImageType())
                .setFileVersion(otaFile.getFileVersion())
                .setFileOffset(fileOffset)
                .setImageData(imageData));
```

I think that this provides a more scalable solution – ie we can also append ```setDefaultResponse()``` or ```setApsAck()``` on this where going through the existing cluster methods with the fixes parameters is not so simple.

I’d welcome your thoughts, but I think that will simplify things and if you don’t foresee any issues I will take a look over the next few days at refactoring the autocoder to do this (the changes are pretty small so it won’t take a load of time).

wsowa:
By asking about the autocoder I was more thinking "why do you have cluster's spec in XML at all" they could be just a regular Java code. I would see the value in interpreting at runtime or transforming at (pre-)compile time the XMLs if they would be provided externally by some standardization body like zigbee alliance. That way you would have reliable source of protocol spec that your library automatically implements whenever an official specification is updated. This is not the case though. The generator, schema and XMLs are your contribution if I understand correctly so you effectively implement Zcl classes in the XML manually rather than in Java. I appreciate that there's a lot of similarity between clusters but there's nothing better in copying and modifying XML descriptors then copying and modifying Java classes code and a lot of code duplication can be eliminated with inherence and composition. Also XML is less extensible then java. You already have special autocoder logic for ZDO and my commit above was another attempt to add to the autocoder something that would be naturally easy in Java.

With your changes in #948 many of the cluster classes will have no non-deprecated methods so you could consider removing these cluster classes completely. However I would suggest to keep them and use java to improve quality of framework that they currently provide by making more use of static type checking. E.g. currently you can sent OnOffCommand to the Ota cluster. No compile time check. But if the ZclCluster.send* methods would accept a generic argument that extends a ZclCommand rather then just ZclCommand then the OnOffCluster could overwrite them and accept only subclasses of abstract OnOffClusterCommand and similarly Ota Cluster could overwrite them to accept only subclasses of abstract OtaClusterCommand. Going further in the Command classes you could invert the control a bit and have a sendResponse method that would accept the cluster of correct type as a context and reply to this particular request. That would allow to make sure that you can response to a command only with responses that are actually responses that protocol specifies rather than any response from the same cluster. Again something that would be difficult to define in XML but fairly easy in Java. Testing is another aspect. If a lot of cluster code could be abstracted then testing those abstractions would cover quite a lot of Zcl estate.

I have commented separately in the #948 your change proposal. I think it is OK with exception to the risk you're creating by allowing to construct Commands empty (without arguments) so if you don't like above suggestions or you find then too radical to implement now then please carry on with #948 

Best Regards,
Witek

cdjackson:
@wsowa many thanks for your review and also your thoughts - I appreciate the feedback, even if we may not agree on XML, I will try and address your points, and maybe I can change your mind (at least a little ;) ).

> By asking about the autocoder I was more thinking "why do you have cluster's spec in XML at all" they could be just a regular Java code.
> 
Defining in XML reduces a lot of duplication in the code which reduces potential bugs and improves quality. It also makes it a LOT easier to make more substantial changes (as we have done here) - for example with deprecating methods, and is a LOT easier to review and fix issues than Java.  To explain this statement, if I have 1 parameter in a command, this generates at least 5 pieces of code from this one parameter definition in XML - there is the parameter definition, a getter and setter, and a serialise/deserialize method. If we make a mistake in the XML, then it gets fixed in all 5 places with 1 change in the XML and everything is consistent. For commands with more parameters (some commands may have 5 or 10 parameters) we have to generate all this code, and in the serialise/deserialise methods, everything has to be correct, and in the right order. Again, doing this in the XML is a lot easier and ensures that the Java is both methods is correct. 

The XML format is quite similar to the tables that are in the ZigBee docs, so it’s pretty easy to review, and the parameters in the ZigBee docs can easily be copied into the XML which generates multiple parts of Java code with one line of XML.


> I would see the value in interpreting at runtime or transforming at (pre-)compile time the XMLs
> 
I did think about this, but this has the downside that for debugging we don’t have the concrete classes so I considered it more problematic.


> if they would be provided externally by some standardization body like zigbee alliance.
> 
I agree, but ZigBee don’t provide this unfortunately. ZWave do have such a thing. There are other similar things done by (for example) Silabs who also have their own XML format and I do have the ability to cross check against their XML.


> there's nothing better in copying and modifying XML descriptors then copying and modifying Java classes code and a lot of code duplication can be eliminated with inherence and composition.
> 
Again, that’s not correct as we define parameters and attributes once in the XML, and these are used in multiple places. If there’s an error in the XML, it is fixed once and propagated throughout the Java.

Anyway, I’m afraid that I will not be changing this to remove the autocoder and you will have to trust me that it has been totally invaluable over the years in keeping a consistent codebase that is easily maintainable.

> With your changes in #948 <https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/948> many of the cluster classes will have no non-deprecated methods so you could consider removing these cluster classes completely.
> 
Methods are only one part of the clusters - they also provide all the attributes of a cluster. So, yes, if there are no attributes then what you say is correct, but clusters DO have attributes.


> if the ZclCluster.send* methods would accept a generic argument that extends a ZclCommand rather then just ZclCommand then the OnOffCluster could overwrite them and accept only subclasses of abstract OnOffClusterCommand
> 
Nice idea - I do like this.


> Command classes you could invert the control a bit and have a sendResponse method that would accept the cluster of correct type as a context and reply to this particular request.
> 
I’m not sure I completely follow this. How can you be sure that the response is consistent with the specific request? (At least without having a table of allowable responses to any request, which is possible of course, but quite an extension (but maybe still a good idea to add in time).

> Again something that would be difficult to define in XML but fairly easy in Java.
> 
I don’t think so - it’s easy to define this in XML - just as you already did - you simply add AllowableResponse=“XYZ” to the XML.


Thanks for your thoughts - I really do appreciate it. I think we have a different view about the advantage of the XML - while I do agree we loose some flexibility, it has really been invaluable over the years, and I believe ensures a better quality product as it guarantees consistency throughout the code which would be difficult to achieve otherwise. I’m afraid I won’t be changing that ;)

I would like to explore your ideas further if you like - maybe the best thing is to discuss them in separate issues?

Thanks.

wsowa:
Hi,
Thanks for taking time and sharing your point of view. Appreciated. I am less precious about XML thing then static type checking. I still do feel that we could get to similar level of reduction of code duplication just with Java but I have no arguments to prove it other then trying it and that is not a quick test to run so I will pass and give my thumbs up to XMLs.

I will clarify my idea on responses on the OTA example...
You could have a sendResponse() method in QueryNextImageCommand that takes a cluster as an argument and all data necessary to construct a response the method's signature would be more or less:
```
class QueryNextImageCommand extends ZclCommand {
  ...
  void sendResponse(ZclOtaUpgradeCluster cluster, ZclStatus status, Integer manufacturerCode, Integer imageType, Integer fileVersion, Integer imageSize) {
    // send from cluster QueryNextImageResponse matching this command
  }
}
```
This way you we would guarantee that a) response is always sent to a correct request b) response is always sent to a correct cluster. With cases like Page Image Request where the first response is treated differently (requires acks) than non-first you could put that first vs non-first logic into a request class. What it turn makes it easier to code a multiple requests being processed simultaneously.
Not sure how useful would that be with other clusters though...

PR #940, opened 2019-12-30
Fix updating transaction state for (near to) fragmented frames

wsowa:
When I was working on a test to demonstrate the problem I tried to fix with one of the commits in the PR #933 I found out that my fix was actually solving the problem only for case where APS payload length == fragmentation length. The original fix didn't address issues with payload > fragmentation threshold.

This PR attempts to fix it. The problem in general was that both TX and RX ACK were accounted for a frame fragment (block as called in ZigBee Specification) receipt while only RX ACK should. That lead to premature passing of RX/TX Ack to transaction manager (for payload > fragmentation length) OR not passing RX Ack to transaction manager (for payload = fragmentation length which was my initial case).

With this fix TX/RX Ack is passed upwards only for the last block being TX/RX Acked. There are two assumptions made here. One is that TX Ack arrives always before RX Ack. Other is that each block is RX Acked. Both are true as long as the implementation keeps only one block in the flight (fragmentation window is effectively = 1) what is the case at the moment. If this is going to change than we will need more advances tracking of each of the blocks state anyway as they can get ACKed out of order and more than once each.

I suggest to review the four tests I've added first. I think they express the intent quite clearly but obviously the implementation change is irrelevant if the testcases are incorrect. Running these tests on current implementation to see what the problem is.

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/aps/ZigBeeApsFrame.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/aps/ZigBeeApsFrame.java
@@ -426,4 +426,23 @@ public class ZigBeeApsFrame {
         return builder.toString();
     }
 
+    /**
+     * Calling this method indicates that transmission of a fragment has completed.
+     * It moves the fragment base and decrease outstanding fragments counter.
+     */
+    public void oneFragmentCompleted() {
+        fragmentBase++;
+        if (fragmentOutstanding > 0) {
+            fragmentOutstanding--;
+        }
+    }
+
+    /**
+     * Calling this method indicates that a fragment has been sent. It increases outstanding fragments counter.
+     */
+    public void oneFragmentSent() {
+        if (fragmentOutstanding <= fragmentTotal) {
+            this.fragmentOutstanding++;
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/aps/ZigBeeApsFrame.java
```diff
@@ -426,4 +426,16 @@ public String toString() {
         return builder.toString();
     }
 
+    public void oneFragmentCompleted() {
```
cdjackson: Please can you add a line or two of Javadoc to this since it's a public method (I know this class isn't the best example, but I like to try and have the documentation.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/aps/ZigBeeApsFrame.java
```diff
@@ -426,4 +426,16 @@ public String toString() {
...
+        fragmentBase++;
+        if (fragmentOutstanding > 0) {
+            fragmentOutstanding--;
+        }
+    }
+
+    public void oneFragmentSent() {
```
cdjackson: Please can you add a line or two of Javadoc to this since it's a public method

cdjackson:
Thanks for this @wsowa - I think it looks fine. I will run a few tests with this before merging - I don't see any problem, but just want to check. Probably I will do this tomorrow.

(I've just got a couple of small comments).

cdjackson:
Thanks @wsowa this looks fine and initial testing looks good. I think the assumptions are ok (window size of 1) for now.

I will merge this and perform some further testing within my system.

PR #948, opened 2020-01-06
Ota upgrade

cdjackson:
This PR includes two commits -:
* Refactoring of the ```ZclCluster``` and derived classes to improve the sending of commands and responses.
* Modifying the ```ZclOtaUpgradeServer``` as per the changes in #936 from @wsowa but using the refactoring in the above commit.

The refactoring in the first commit makes both ```sendCommand``` and ```sendResponse``` public methods in ```ZclCluster```. It deprecates all the current methods used to send commands in the ```ZclCluster``` class in favour of a ```Builder``` style notation used to generate and send commands (as seen in ```ZclOtaUpgradeServer```). This provides more flexibility in that we can not add additional methods to change the instantiation of the command (eg by disabling the default response).

Probably will replace #936 - @wsowa it would really be great if you could take a look. I think I've re-created the functionality that you provided in #936 and I successfully tested an OTA transfer to an OSRAM bulb. There may be something still wrong with fragmentation, but this was in any case set too low so I've increased the fragmentation limit to avoid this for now.

Also-by: Witold Sowa <witold.sowa@gmail.com>


cdjackson:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/ImagePageCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/MatchDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/LocationDataResponse.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/ViewSceneResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/NodeDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/EndDeviceBindRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/DeviceConfigurationResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/RestartDeviceResponseResponse.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/general/WriteAttributesCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/SetAbsoluteLocationCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ManagementLqiRequest.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpProxyTableRequest.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishConsolidatedBillCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/RestoreStartupParametersResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpPairing.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpTranslationTableUpdate.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishBlockThresholdsCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/ResetStartupParametersResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/messaging/DisplayMessageCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/QuerySpecificFileResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/LockDoorResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ManagementRoutingRequest.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/RemoveAllScenesCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ActiveEndpointsRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/QueryNextImageCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ComplexDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/GetBlockThresholdsCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/ToggleResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/UnlockWithTimeoutResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/SaveStartupParametersResponse.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/QueryNextImageResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/colorcontrol/EnhancedStepHueCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/UnlockDoorResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/colorcontrol/StepHueCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpPairingSearch.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/ImageBlockResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/AddSceneCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpCommissioningNotification.java  -4
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/groups/RemoveGroupCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/general/WriteAttributesNoResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/EnhancedAddSceneCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/LockDoorCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/GetSceneMembershipCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/GetPriceMatrixCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/LocationDataNotificationCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/levelcontrol/StepWithOnOffCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/PowerDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/messaging/GetLastMessage.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ManagementBindRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpTunnelingStop.java  -4
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/EnhancedViewSceneResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/ImageBlockCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/groups/ViewGroupCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/UnlockDoorCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/UserDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpSinkTableRequest.java  -4
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/GetTierLabelsCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/general/WriteAttributesUndividedCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishBillingPeriodCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/SetDeviceConfigurationCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishPriceMatrixCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/colorcontrol/StepSaturationCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/levelcontrol/StepCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/messaging/DisplayProtectedMessageCommand.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/zsmartsystems/com-zsmartsystems-zigbee/pullRequest?prid=4772145&bid=15902271)

cdjackson:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/PowerDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/UnlockDoorResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/RemoveAllScenesCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishBillingPeriodCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/QueryNextImageResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/SetAbsoluteLocationCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/AddSceneCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/DeviceConfigurationResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/ResetStartupParametersResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/ImagePageCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ComplexDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/groups/RemoveGroupCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/ViewSceneResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/LockDoorResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/EndDeviceBindRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/RestoreStartupParametersResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/UserDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpTranslationTableUpdate.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/colorcontrol/StepHueCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/ImageBlockCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishConsolidatedBillCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/groups/ViewGroupCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/QueryNextImageCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/UnlockDoorCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/levelcontrol/StepWithOnOffCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/LocationDataResponse.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/general/WriteAttributesNoResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/MatchDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/ToggleResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/NodeDescriptorRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/LocationDataNotificationCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ActiveEndpointsRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpPairing.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpPairingSearch.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpCommissioningNotification.java  -4
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpSinkTableRequest.java  -4
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/general/WriteAttributesCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/GetTierLabelsCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/colorcontrol/StepSaturationCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/rssilocation/SetDeviceConfigurationCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/UnlockWithTimeoutResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpTunnelingStop.java  -4
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/GetPriceMatrixCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/general/WriteAttributesUndividedCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishBlockThresholdsCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/levelcontrol/StepCommand.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ManagementBindRequest.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/colorcontrol/EnhancedStepHueCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/ImageBlockResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/greenpower/GpProxyTableRequest.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/messaging/DisplayMessageCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/PublishPriceMatrixCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/messaging/DisplayProtectedMessageCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/price/GetBlockThresholdsCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ManagementLqiRequest.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/SaveStartupParametersResponse.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zdo/command/ManagementRoutingRequest.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/doorlock/LockDoorCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/GetSceneMembershipCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/messaging/GetLastMessage.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/EnhancedViewSceneResponse.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/commissioning/RestartDeviceResponseResponse.java  -2
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/scenes/EnhancedAddSceneCommand.java  -1
+ com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/clusters/otaupgrade/QuerySpecificFileResponse.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/zsmartsystems/com-zsmartsystems-zigbee/pullRequest?prid=4772145&bid=15902271)

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/aps/ApsDataEntity.java
```diff
@@ -45,7 +45,7 @@
 public class ApsDataEntity {
     private static final long DUPLICATE_TIME_WINDOW = 5000;
 
-    private static final int FRAGMENTATION_LENGTH = 65;
+    private static final int FRAGMENTATION_LENGTH = 80;
```
wsowa: Why 80?

cdjackson:
As mentioned, there still seems to be an issue with fragmentation, and I wanted to avoid fragmentation by moving the fragmentation length above the frame length (which was 78 if I remember correctly). This needs to be resolved of course, but I wanted to make this PR available for comment before I left.

com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeZclClusterGenerator.java
```diff
@@ -397,6 +397,7 @@ private void generateZclClusterClasses(ZigBeeXmlCluster cluster, String packageR
 
             out.println("     * @return the {@link Future<CommandResult>} command result future");
             out.println("     */");
+            out.println("    @Deprecated");
```
wsowa: Should we have a statement in the Javadoc like following?
`@Deprecated use generic sendCommand/sendResponse methods in ZclCluster instead`

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
```diff
@@ -31,6 +31,7 @@
 import com.zsmartsystems.zigbee.zcl.clusters.ZclOtaUpgradeCluster;
 import com.zsmartsystems.zigbee.zcl.clusters.otaupgrade.ImageBlockCommand;
```
wsowa: Should we call it `Image Block Request`? Both requests and responses are commands. When we call it Image Block Command it is not clear whether it is a response or a request.
Same applies to other OTA Commands that are Requests:
Image Block Request
Image Page Request
Upgrade End Request
Query Next Image Request
Query Specific File Request

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
```diff
@@ -514,40 +524,55 @@ public void run() {
...
-    private int sendImageBlock(int fileOffset, int maximumDataSize) {
+    private int sendImageBlock(ZclCommand command, int fileOffset, int maximumDataSize) {
         ByteArray imageData = otaFile.getImageData(fileOffset, Math.min(dataSize, maximumDataSize));
         logger.debug("{} OTA Data: Sending {} bytes at offset {}", cluster.getZigBeeAddress(), imageData.size(),
                 fileOffset);
-        cluster.imageBlockResponse(ZclStatus.SUCCESS, otaFile.getManufacturerCode(), otaFile.getImageType(),
-                otaFile.getFileVersion(), fileOffset, imageData);
+
+        if (command == null) {
+            cluster.sendCommand(new ImageBlockResponse()
```
wsowa: Can we extract
```
new ImageBlockResponse()
                    .setStatus(ZclStatus.SUCCESS)
                    .setManufacturerCode(otaFile.getManufacturerCode())
                    .setImageType(otaFile.getImageType())
                    .setFileVersion(otaFile.getFileVersion())
                    .setFileOffset(fileOffset)
                    .setImageData(imageData)
```
to a variable before the if and then within if/else only call setDisableDefaultResponse and use proper sending method? Or get the setDisableDefaultResponse outside the if and call it with `cluster==null` arg
o

wsowa:
Hi @cdjackson 
I like the direction where it is going, in particular exposing sendCommand method to the application. Interestingly some of the autogenerated Clusters classes will consist only of deprecated public methods now. That brings up the question I have asked on the other day about why do you autogenerate the code in the first place...

I think the OTA related changes are ok but I will be able to test them later this month so I suggest to merge it now and then if I find any issues in tests I'll raise a separate PR.

With these change we now have lots of places where we call deprecated method. I appreciate it is a tedious work but would you update all these calls now or rather consequently one by one when you touch that code?

I do have a bit of a problem with the builder approach. I appreciate it is more extensible but on the other hand it allows one to construct an invalid command (i.e. with some mandatory attributes not set) without any compile time checks so it will inevitably bring some bugs over time when someone will forget to call a setter. This is one of the reason why I would hesitate to refactor all the calls of all the deprecated methods as it could easily introduce bugs that will be hard to catch given not perfect tests coverage. I would love a mixed solution where you have (the only) constructor with all the required command fields as attributes that could be then supplemented with chained setters to add optional attributes (like disable response or disable aps ack or a command-specific optional field). Chained setters wouldn't be necessary then as in most of the cases you usually want to populate a command with all the required fields and nothing else. For other cases a one or two attributes could be set traditionally without chaining successfully. This way your setters would stay regular void setters.

wsowa:
Ok... I don't think it can be merged into master and certainly not released until this is resolved as with this change if I understand correctly we may attempt to transmit frames that are too long and therfore won't be transmitted at all.

wsowa:
Ok... I don't think it can be merged into master and certainly not released until this is resolved as with this change if I understand correctly we may attempt to transmit frames that are too long and therfore won't be transmitted at all.
> As mentioned, there still seems to be an issue with fragmentation, and I wanted to avoid fragmentation by moving the fragmentation length above the frame length (which was 78 if I remember correctly). This needs to be resolved of course, but I wanted to make this PR available for comment before I left.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
cdjackson: > Should we call it ImageBlockRequest

Possibly it would be better, however for historical reasons this has always been the name and I don't want to get into major refactoring.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
cdjackson: Yes, Good point

cdjackson:
Currently this should not require fragmentation - these frames should be (and are being) sent fine without fragmentation, so we need to set the fragmentation level at a point that avoids needlessly fragmenting frames (which this does). If there is an issue with fragmentation (which needs checking) then it should not be impacted with this change to the way commands are sent generally, and in OTA.

cdjackson:
> some of the autogenerated Clusters classes will consist only of deprecated public methods now. 

Yes, but clusters do not only contain commands, they also contain the lists of attributes that get maintained to retain the state of the remote device. There of course may be some clusters that have no attributes, but probably it is not so many. As you've also pointed out elsewhere, by subclassing commands we can ensure the cluster only accepts commands that are meant to be sent through that cluster, which I like and will look at implementing.

cdjackson:
> I do have a bit of a problem with the builder approach. I appreciate it is more extensible but on the other hand it allows one to construct an invalid command (i.e. with some mandatory attributes not set) 

Fair point.

This could probably be resolved by adding a check in the serialiser - although I suspect what will happen if you do not set everything is you will get an exception thrown saying "Unable to serialise null value" - or something like that. So I think at runtime this will be caught now, even if it could probably be handled better.

One advantage of deprecating these methods is that some classes (especially int he Smart Energy area) have hundreds of commands - often with many parameters. This has generated huge cluster classes which is a downside of the XML in some ways, but on the other hand it is consistent across the framework. Changing the approach to use the setters would significantly improve those classes - especially given that most applications will not use more than a few of the methods.

I personally like the builder approach as I do think it's cleaner, and probably generates more readable code (although I would agree that is likely to be personal preference). I agree there's nothing to stop invalid commands being produced at compile time, but I think it is caught at run time, and this could probably be improved.

If we improved the error checking, would you be happier with the builder approach? 

> I would love a mixed solution where you have (the only) constructor with all the required command fields as attributes that could be then supplemented with chained setters to add optional attributes 

We could of course do both - ie provide the constructor with all the fields, and provide the setters that return the object. This does provide more flexibility to the user, although I'm a bit concerned that choice is not always a good thing ;)

Happy to hear your thoughts on the above points... I'm at the ZigBee test house doing Smart Energy testing at the moment, but will look at this further later and may implement your subclass idea as part of this refactoring (it's simple with the autocoder to change the class a child class extends ;) ).

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
wsowa: Ok

wsowa:
It's true that incorrectly constructed command will fail in serialization but this is a run time check so not really great. Improving error checking in the runtime would help with readibility of error (nice log/exception instead of NPE) but is still a run time check so no much improvement. I really prefer the code to validate at compile time because in practice with modern IDEs like IntelliJ or even Eclipse it is a writing time check. The quickest feedback loop you can get. I agree that some things are personal preferences and I try to not discuss such but here we have two comparable solutions one with runtime checking and the other with compile time checking. It is a good thing to have objects guaranteed to always be valid business-wise. 

I am less precious about XML thing but static checks are valuable and easy to get In that case.
I will post screenshots of my IDE with both options to compare clarity of code later. Do you have a candidate command that you would like to see in you mind? A complex one would be good test subject.

cdjackson:
While it’s true that these are runtime checks, if the methods are incorrectly called (or not called) the developer will see them pretty quickly. It’s impossible to completely ensure the developer is doing the right thing - they can still pass invalid parameters into a method. (But ok, I agree it reduces the chances and that’s a good thing ;) ).

I’ll take another look at this later in the week when I get a bit more time - maybe the mixed idea (constructor with params) is ok.

Any command is fine as a demo - thanks.

wsowa:
I took PublishPriceCommand as an example as it has lots of arguments and half of them are optional.
Here's comparison of sending this command with all mandatory arguments plus two optional
Screenshot_20200108_222725

Here is the same but with one of the mandatory arguments missing due to developers error
Screenshot_20200108_222708

The clarify difference in the first case is none in my opinion but that's matter of style/aesthetics so some people may like one more then the other.

The second case though demonstrate the power of the solution with constructor arguments as it highlights the error immediately.

Please take a look and make a call I won't be advocating for this more.

cdjackson:
Thanks for the useful discussion @wsowa.

I suspect you picked an interesting command to show here as most commands don't have optional parameters like this, and we don't currently have this as a concept. It could of course be added, but for now I don't see the need (we are not currently using the price cluster in our current Smart Energy product).

Anyway, I think on balance your suggestion works best and is safest, so I will go with that.

What I will do is to close this PR and implement the changes in 3 PRs for visibility - since the changes impact many files I prefer not to mix things up. So, I think there will be 3 PRs -:

* Refactor the method ```send``` to ```sendCommand``` and make it ```public```
* Deprecate the methods in the ```ZclCluster``` classes and add the new constructor (with parameters) to the ```ZclCommand``` classes
* Make the changes to the OTA server

wsowa:
Yeah, I've chosen quite complicated command intentionally. For simple ones the benefits are still there though. 

Given it is quite tedious and error prone process to identify what command arguments are mandatory and what are optional and that non present optional argument needs to be serialized in a argument specific way (which would need to be defined in the XML too) I suggest that for now we keep them all as mandatory. This has a nice side effect that all Command fields will be initiated in the constructor so we can declare them all final and remove all the setters making the class immutable in effect. Which is a good property of a DTO. 

If in the future you decide to represent optional arguments in the API (if ever, I'm not sure how many commands have such) then I suggest you do so wrapping them in Opional<>. The user of the API will need to explicitly use Optional.none() if he decides to not use the optional argument. Which again is good think as makes the intention explicit.

cdjackson:
Absolutely - that’s exactly what I was going to do :)

If I need to use (for example) the Price commands, then I might consider the optional idea as you’ve suggested - otherwise we’ll just use the builder setters for things like the default response. There are not so many commands that have optional parameters…

PR #955, opened 2020-01-10
Refactor ZclCluster.send() to sendCommand() and provide public methods

cdjackson:
As per the discussion in #948, this is the first part in refactoring the sending of commands through ZclCluster. This PR should not change any external interface, but will require users to recompile.

This provides a public ```sendCommand``` method alongside the existing ```sendResponse``` method. Both commands are now implemented in the cluster implementations to enforce that only commands for a specific cluster may be sent through that cluster. This changes the class hierarchy, but should remain source compatible to users.

The existing protected ```ZclCluster.send()``` method has been refactored to ```ZclCluster.sendCommand``` to be consistent with the above.

All commands now extend from an abstract base class specific to the cluster to enforce type checking of commands.

@wsowa please let me know if you have any comments - this implements what I understand from your suggestion in the previous thread. (note that deprecating the cluster commands and adding command constructors with the parameters will be a subsequent PR.

### DIFF
--- a/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeBaseClassGenerator.java
+++ b/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeBaseClassGenerator.java
@@ -83,6 +83,7 @@ public abstract class ZigBeeBaseClassGenerator {
         fixedCaseAcronyms.add("MAC");
         fixedCaseAcronyms.add("MMO");
         fixedCaseAcronyms.add("NWK");
+        fixedCaseAcronyms.add("OTA");
         fixedCaseAcronyms.add("PIN");
         fixedCaseAcronyms.add("PIR");
         fixedCaseAcronyms.add("RMS");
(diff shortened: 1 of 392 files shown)


wsowa:
Looks good to me. One thing that I would consider is to make ZclCluster a generic class, i.e. declare it as
`public abstract class ZclCluster<C extends ZclCommand>` and use C as an argument type to public sendCommand and sendResponse and internally in the ZclCluster implementation as well where possible. Then the particular clusters would be declared as 
`public class ZclOnOffCluster extends ZclCluster<ZclOnOffCommand>` and there would be no need to overwrite sendCommand/Response methods in all the clusters. That would safe two generated methods per cluster.

cdjackson:
Thanks - but did you actually try this to see if it works?

wsowa:
I did draft it. It becomes quite messy internally in ZclCluster. Requires
some amount of casting to match generic methods definitions or possibly
more major refactoring would make it easier. I think we should keep it as
it is for now. The benefits it bring are not worth the mess it will bring
to ZclCluster or effort of the refactoring.

No other comments from my side. Looks good to merge.

W dniu sob., 11.01.2020 o 08:47 Chris Jackson <notifications@github.com>
napisał(a):

> Thanks - but did you actually try this to see if it works?
>

cdjackson:
Ok, thanks. That’s also what I found but I was wondering if maybe you’d found a better way through and come to a different conclusion.

PR #968, opened 2020-01-15
Catch uncaught exceptions in attribute listeners and log them

### LINKED ISSUE #967: Exceptions in attribute listeners are not caught by the framework
The method `ZclCluster.notifyAttributeListener` notifies all registered `ZclAttributeListener`s when an attribute is updated.

If one of the listener throws an exception, the exception is not handled in `ZclCluster.notifyAttributeListener`, with two effects:
1. The attribute listeners following the listener that threw the exception will not be notified about the attribute update.
2. The exception is not logged with the used logging framework, but is logged to the console instead.

My suggestion is to 
* (a) handle exceptions in `ZclCluster.notifyAttributeListener` by simply logging them using the logger of the `ZclCluster` class, and to 
* (b) add an `UncaughtExceptionHandler` to the threads that are created by the executors provided by the `ZigBeeExecutors` class, which also simply logs the exception.

I will provide a PR shortly, as a basis for discussion.

hsudbrock:
Also adds an UncaughtExceptionHandler that logs exceptions to the executors
provided by the framework.

Resolves #967

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeExecutors.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeExecutors.java
@@ -24,8 +24,8 @@ import org.slf4j.LoggerFactory;
  *
  */
 public class ZigBeeExecutors {
-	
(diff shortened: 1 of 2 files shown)


hsudbrock:
Thanks Chris, I have run the code formatter.

cdjackson:
Thanks Henning.

PR #974, opened 2020-01-27
Fix ClassCastException in console command for reading reporting configuration

### LINKED ISSUE #973: ClassCastException in console command for reading reporting configuration
The console command implementation in `ZigBeeConsoleReportingConfigCommand` sends a `ReadReportingConfigurationCommand` to a device, but then tries to get the response as a `ReadAttributesResponse`. This has led to a `ClassCastException` when I tried to use that console command, because the response is actually a `ReadReportingConfigurationResponse`. ([see here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/5a5719275b6cac929f9fde27ff6e9d32f7489a5e/com.zsmartsystems.zigbee.console/src/main/java/com/zsmartsystems/zigbee/console/ZigBeeConsoleReportingConfigCommand.java#L70))

Do I overlook something simple here? (From the git history, it seems to have been like this like forever, and I wonder whether I do something wrong, or whether this small issue just went under the radar for all that time.)

Anyways, I will attempt a PR for fixing this.

hsudbrock:
In the console command for reading reporting configuration, evaluate the response as a `ReadReportingConfigurationResponse` (and not as a `ReadAttributesResponse`).

Fixes #973


cdjackson:
Thanks Henning. I guess this was a cut and paste from a previous console command so I don't think you overlook anything.

PR #985, opened 2020-02-17
Fix checksum error in FIND_DEVICE_CONFIRM on CC2531 and improve logging

wsowa:
It's a minor fix that I made while debugging issue described in https://community.openhab.org/t/zigbee-devices-work-for-sometime-and-then-suddenly-stop-working/92231/52 that I thought would be good to keep.
It doesn't correct any significant behavior but adds a bit more logging and fix parsing of a packet that is ignored anyway but at least we will see no exception on the console.


cdjackson:
Thanks @wsowa

PR #987, opened 2020-02-18
Fix deadlock in CC2531 between sending and input parsing threads

wsowa:
Branches from 1.2.9...


cdjackson:
Why is this based against 1.2.9? Presumably if we rebase this will be fine.

wsowa:
Yes, the change boils down to wrapping a block of code in a new thread so shuld be easy although not conflict less to rebase. 

I hoped that branching from 1.2.9 will make it easier to relaese hot fix for openhab 2.5.1 binding. Knowing now this is not the case I am happy to re issue the PR. Question is where to branch from? Master is currently at 1.3.0 and I don't know if you plan to switch over to 1.3.x in OH 2.5.2 or do you have more changes planned before releasing 1.3.0 and 2.5.2 will stay on 1.2.x with new 1.2.10 version including this change. 

Feel free to rebase or let me know where to branch from. My only goal is to bring this change to OH as soon as possible

wsowa:
Ok. Rebased to master

cdjackson:
Thanks @wsowa

PR #990, opened 2020-02-21
Replacing the PrinterWriter with PrinterStream to make autocode UTF-8 safe

bluelineXY:
As on some non UTF-8 Systems like windows the generated files are not compileable.

Please handle this as Proposal

Signed-off-by: bluelineXY <5988879+bluelineXY@users.noreply.github.com>

### DIFF
--- a/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeBaseClassGenerator.java
+++ b/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeBaseClassGenerator.java
@@ -13,7 +13,8 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.PrintWriter;
+import java.io.PrintStream;
+import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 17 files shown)


cdjackson:
@bluelineXY can you provide an example of what isn't compilable? I assume something coming through from some XML files is causing a problem?

bluelineXY:
Hi, 
have use the autocode generator on Windows 10 for example. 
The issue occurs if the Documentation for example contains "´" as this will be transformed wrong. 

In addition to this it makes sense to ensure "UTF-8" generated files using autocode as the UTF-8 Format ist defined in the pom.xml files as expected encoding.

cdjackson:
Ok, but can you provide an example? I'm trying to understand the extent of the issue

bluelineXY:
Hi, 
as example: 
```
File: /com/zsmartsystems/zigbee/zdo/command/UnbindRequest.java
Two files have different content. They differ at line 66
File1 has      * 64-bit extended address for DstAddress and DstEndp present 0x04 ▒ 0xff = reserved and File2 has      * 64-bit extended address for DstAddress and DstEndp present 0x04 – 0xff = reserved at line 66
```
The issue is between 0x04 and 0xff. 
Using as proposed the Stream it ends up like expected: 

```
File: /com/zsmartsystems/zigbee/zdo/command/UnbindRequest.java
Two files have same content. 
```

cdjackson:
Thanks @bluelineXY - I don't see any real downside to this so it seems a safer approach and it's limited to the autocoders anyway.

Issue #989, opened 2020-02-21
Storage of ZCL attribute datatypes by the network data store


hsudbrock:
Currently, implementors of `ZigBeeNetworkDataStore` need to store information about ZCL attributes like the datatype or whether the attribute is mandatory. When reading a node from the data store, new instances of `ZclAttribute` are populated with that data, and those new instances are then used instead of the attributes defined in `ZclCluster.initialize{Server|Client}Attributes`.

I wonder if this should really be done that way. The method `ZclCluster.initialize{Server|Client}Attributes` provides the values defined in the ZCL spec, and I do not see why it should be possible that implementors of `ZigBeeNetworkDataStore` should be able to provide different values for those attribute properties.

Why do I even care about this? If there is, at some point in time, wrong attribute information in the autocoder XML files, this wrong information will make it into the data written by a network data store. If the autocoder XML files are fixed, this fix will never make it into the `ZclAttribute` instances that are created from network data store information that was written before that fix.

One concrete example is #534, which fixed wrong datatypes in the thermostat cluster XML files. I have one system here which reads a thermostat node that was persisted before #534 was merged, and in consequence this node still contains wrong attribute datatypes.

Chris, do you think it is sensible to  not create new instances of `ZclAttribute` from a `ZclAttributeDao` [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/ef8abce88cb3328052d44a987b29aa8e6a342e3f/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java#L1591) in `ZclCluster.setDao`, but rather to update the attributes created by `initialize{Server|Client}Attributes` during object creation with the infos from the `ZclAttributeDao` (excluding information defined by the ZCL spec like datatype, mandatory, etc.)?

cdjackson:
In general, I don't mind this approach and can certainly see some benefit. However, there are a couple of things that we should keep in mind.

1) Custom attributes. Unless we can be sure that any custom attributes are added to the cluster before the data store updates it, then you will loose them. Since the ```ZclCluster``` is instantiated as part of the deserialisation from the data store, I don't think we can account for an manufacturer specific attribute?
2) Attribute "attributes" would be lost. By this I mean the time of last update, and the last value, or if it is even supported (see also #931).

I'm not completely sure we can overcome these two issues (at least the first) if you are only updating existing attributes.

[...]

PR #994, opened 2020-03-04
Add the manufacturer code to the ZclAttributeDao

hsudbrock:
This field has previously been missing in the `ZclAttributeDao`, but should be stored
in the database as well.

What does go wrong without this fix? After re-filling the node list from the database, operations like, e.g., attribute reads for manufacturer-specific attributes will fail because the manufacturer-specific bit will not be set in the requests anymore.

Chris, this is one of the issues that we discussed in #989.


cdjackson:
Thanks Henning.

PR #995, opened 2020-03-05
Fix unit test for receive default command

hsudbrock:
The unit test `ZigBeeNetworkManagerTest.testReceiveDefaultCommand` should originally test that the `ZigBeeNetworkManager` does not notify about default responses. However, after ignoring endpoints other than 1 (cf. https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/829) the test does not test this anymore. Why is that? The test uses a default response to endpoint 2, and this message is now dropped not because it is a default response, but because it is not for endpoint 1.

This PR adapts the unit test to use a default response for endpoint 1, so that it actually tests the correct behaviour for default responses again. It slightly extends the mocking, because otherwise the test runs into a NPE (this was not necessary when the test was first implemented, but due to other changes in the `ZigBeeNetworkManager` this  is now necessary...)

### DIFF
--- a/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/ZigBeeNetworkManagerTest.java
+++ b/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/ZigBeeNetworkManagerTest.java
@@ -476,6 +476,7 @@ public class ZigBeeNetworkManagerTest
 
         ZigBeeEndpoint endpoint = Mockito.mock(ZigBeeEndpoint.class);
         ZclCluster cluster = new ZclOnOffCluster(endpoint);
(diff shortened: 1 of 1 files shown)


cdjackson:
> However, after ignoring endpoints other than 1 (cf. #829) the test does not test this anymore. Why is that?

You mean why do we ignore endpoints other than 1? It was required during certification testing to only respond to endpoints that you have registered and are reporting in the descriptors.

cdjackson:
Thanks for spotting this Henning.

hsudbrock:
> However, after ignoring endpoints other than 1 (cf. #829) the test does not test this anymore. Why is that?

Actually, I meant it as "Why does the test does not test it anymore?", giving the answer afterwards myself. But your answer why endpoints other than 1 are ignored was interesting for me nonetheless. Actually, I stumbled upon the "broken" test when playing around with allowing other endpoints :)

Thanks for merging, Chris!

PR #1000, opened 2020-03-12
Use a separate notification service for each network manager

### LINKED ISSUE #999: NotificationService and multiple ZigBeeNetworkManagers
An instance of the `ZigBeeNetworkManager`, in most cases, does not share classes that is uses internally. One exception is the `NotificationService`, which is shared amongst all instances of the `ZigBeeNetworkManager`. 

Currently, this is an issue when using multiple instances of `ZigBeeNetworkManager`: If one instance is shut down via `ZigBeeNetworkManager.shutDown()`, then the executor service of the `NotificationService` is also shut down. In consequence, the other instances won't function correctly anymore, because notifications via the `NotificationService` no longer work.

I think that this is an issue in two cases:
(a) In sytems that handle multiple ZigBee chips in parallel, and
(b) in dynamically updatable systems (like, in my case, an ESH-based system) where a new `ZigBeeNetworkManager` might be already initialized when an old `ZigBeeNetworkManager` is still shutting down. (Unfortunately, ESH does not guarantee that a bridge handler was completely disposed before initializing a new handler for the same bridge.)

I see two possible solutions to improve this:
1) Do not share the `NotificationService` amongst different instances of `ZigBeeNetworkManager`.
2) Properly handle initialization/shutdown in the `NotificationService` even with multiple `ZigBeeNetworkManager`s.

I think that (1) has the advantage that the framework would stop sharing resources between `ZigBeeNetworkManager`s, but the disadvantage that one would have to pass the notification service from the network manager to many objects used by the network manager (transaction manager, clusters, command notifiers, etc.).

I think that (2) has the advantage that it would be a change only in `NotificationService`; my suggestion is to store within the notification service the information which network managers (or alternatively how many network managers) are currently using the notification service, and to shut down the executor service only once all using network managers asked it to shut down.

Do you have an opinion how this could/should be resolved, Chris?

hsudbrock:
As discussed in #999, one option to resolve #999 is to use a separate non-static `NotificationService` for each instance of the `ZigBeeNetworkManager`.

This pull request illustrates how this could look like. The changes are smaller than I had initially thought. Basically, the `ZigBeeNetworkManager` now creates its own instance of the `NotificationService` and provides a method for retrieving the notification manager. Other objects managed by the network mangager (e.g., the transaction manager, the nodes, endpoints, clusters, some applications) get access to the notification service by getting it from their network manager.

Is this what you were thinking of with this solution, Chris?


hsudbrock:
Force-pushed to fix commits (squash the two commits, and sign it correctly) (no code changes).

cdjackson:
Thanks Henning. I guess on balance this is fine - I can't think of a better approach, and this does need to get resolved.

The main reason for delaying my response on this is that you removed the ```ZigBeeNetwork``` interface and I was trying to remind myself why I put this abstraction in in the first place. I don't really remember now and for now it's possibly cleaner.

Can I ask what your thinking was when removing this? I guess you also didn't see the point in it ;)

hsudbrock:
> Can I ask what your thinking was when removing this? I guess you also didn't see the point in it ;)

I was not completely happy removing it, because I think that it is generally a good idea to narrow down the interface to classes that provide a lot of functionality, like the `ZigBeeNetworkManager`.

My first thought was to extend the `ZigBeeNetwork` interface to provide access to the notification service - but I did not like that idea, because providing access to a thread pool does not really fit in semantically to a 'zigbee network' interface.

My second thought was to make the `ZigBeeNetworkManager` implement a second interface that only provides access to the notification service (and to even call that second interface `NotificationService`). But I did not like this idea too much either, because it looked to me like "introducing an interface just to have an interface, not because it makes the code easier to read or maintain" (for instance, the `ZigBeeNode` would then have needed to fields containing the same reference to the `ZigBeeNetworkManager`, one for each interface).

And that was when I thought that in this particular case I did not see so much benefit in having an interface to access the `ZigBeeNetworkManager` from the `ZigBeeNode` at all, and I removed it.

(What I just now saw is that with this, the interface is not used at all anymore, and if we decide to really remove it as suggested in this PR, the complete interface file should be removed; I will add a commit for that after finishing this comment.)

cdjackson:
> What I just now saw is that with this, the interface is not used at all anymore

Yes, this is also what I was looking at yesterday. I agree that if we remove this, then the whole file should be removed.

Ok, I think this approach is probably best on ballance. I'll take one last look later, but I think this is fine.

Thanks for looking into this Henning.

I note that CI seems to be failing. I've just kicked it off again as I see you've just removed the interface while I was typing this, but maybe there is a problem somewhere?

hsudbrock:
Hm, `mvn clean install` works on my machine, but there seems to be a compilation server on the CI server. I will look into it.

hsudbrock:
I will rebase the pull request on `master` to fix the CI build (another PR that was merged a couple of days ago introduced one more static usage of the notification service, and that needs to be adapted to use the notification service non-statically as well...).

cdjackson:
Thanks Henning.

Issue #236, opened 2018-03-12
New attributes not added to cluster follow deserialisation


cdjackson:
If a new attribute is added to the cluster library, then it is not added to the cluster if the cluster is deserialised from storage.

cdjackson:
Closed by #1001

PR #1001, opened 2020-03-13
Only restore attribute state from the node database for non-manufacturer-specific attributes

### LINKED ISSUE #989: Storage of ZCL attribute datatypes by the network data store

hsudbrock:
This resolves #989 and #236.

With this change, when restoring a cluster from a `ZclClusterDao` obtained from the node database, for non-manufacturer-specific attributes only the dynamic attribute state will be restored (i.e., last value, last reporting time, and whether the attribute was marked as implemented).

The benefit is that changes in the attribute definition in the `Zcl***Cluster` classes generated by the autocoder will be applied to attributes, even when the information stored in the node database is from before those changes were made. (E.g., when adding new attributes as described in #236, or when fixing datatypes as described in #989.)

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclAttribute.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclAttribute.java
@@ -543,6 +543,17 @@ public class ZclAttribute {
         reportingTimeout = dao.getReportingTimeout();
         manufacturerCode = dao.getManufacturerCode();
     }
+    
+    /**
+     * Sets the dynamic state of the attribute from a {@link ZclAttributeDao} which has been restored from a persisted state.
+     *
+     * @param dao the {@link ZclAttributeDao} from which to restore the dynamic state
+     */
+    public void setDynamicStateFromDao(ZclAttributeDao dao) {
+        implemented = dao.isImplemented();
+        lastValue = dao.getLastValue();
+        lastReportTime = dao.getLastReportTime();
+    }
 
     /**
      * Returns a Data Acquisition Object for this attribute. This is a clean class recording the state of the primary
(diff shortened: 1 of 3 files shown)


cdjackson:
Thanks Henning.

PR #1004, opened 2020-03-18
Fix auto selecting the quietest channel on Ember

### LINKED ISSUE #1003: Flaw in using the quietest channel for the network (Ember)
The  `EmberNetworkInitialisation` class creates the zigbee network via

```
public void formNetwork(EmberNetworkParameters networkParameters, ZigBeeKey linkKey, ZigBeeKey networkKey) {
.
...
.
    if (networkParameters.getRadioChannel() == 0) {
            networkParameters.setRadioChannel(quietestChannel);
    }
.
...
.
}
```

The problem is, that the `ZigBeeDongleEzsp` which calls this `formNetwork` method, initializes the `networkParameters` during its initialization as follows:

```
// Get the current network parameters so that any configuration updates start from here
networkParameters = ncp.getNetworkParameters().getParameters();
```

As far as I understand it, this means that `networkParameters.getRadioChannel()` will always contain the channel that is set on the chip and thus never 0.

UNLESS one calls `ZigbeeNetworkManager#setZigBeeChannel(ZigBeeChannel channel)` between the `ZigbeeNetworkManager#initialize()` and `ZigbeeNetworkManager#startup()`.

This boils down to the problem of setting a channel that satisfies the check from above: `networkParameters.getRadioChannel() == 0`.

The naive approach with
```
networkManager.setZigBeeChannel(ZigBeeChannel.create(0));
```

results in
```
ZigBeeDongleEzsp] - Unable to set channel outside of 2.4GHz channels: CHANNEL_00
```

The error occurs due to this check in `ZigBeeDongleEzsp`:

```
public ZigBeeStatus setZigBeeChannel(ZigBeeChannel channel) {
        if ((ZigBeeChannelMask.CHANNEL_MASK_2GHZ & channel.getMask()) == 0) {
            logger.debug("Unable to set channel outside of 2.4GHz channels: {}", channel);
            return ZigBeeStatus.INVALID_ARGUMENTS;
        }
        networkParameters.setRadioChannel(channel.getChannel());
        return ZigBeeStatus.SUCCESS;
    }
```

---

As a solution to this problem I think the check in `formNetwork` should be changed to:

```
if (networkParameters.getRadioChannel() == ZigBeeChannel.UNKNOWN.getChannel()) {
            networkParameters.setRadioChannel(quietestChannel);
}
```

And the check for setting a zigbee channel in `ZigbeeDongleEzsp` should be extended to allow for setting the `UNKNOWN` channel, which then means "Use auto detection to find the quietest channel":

```
public ZigBeeStatus setZigBeeChannel(ZigBeeChannel channel) {
        if (channel != ZigBeeChannel.UNKNOWN && (ZigBeeChannelMask.CHANNEL_MASK_2GHZ & channel.getMask()) == 0) {
            logger.debug("Unable to set channel outside of 2.4GHz channels: {}", channel);
            return ZigBeeStatus.INVALID_ARGUMENTS;
        }
        networkParameters.setRadioChannel(channel.getChannel());
        return ZigBeeStatus.SUCCESS;
}
```

However, I am also wondering why there is a check for the 2GHz band in there. Is there any regulatory that Ember chips must only be used within this frequency band?

---

I also had a look at the telegesis implementation, but I think this feature is not implemented (yet), at least I didn't see it in the transport implementation.

triller-telekom:
Using ZigbeeNetworkManager#setZigBeeChannel(ZigBeeChannel.UNKNOWN) before
the networkmanager startup will enable the scan for the quietest channel

Fixes #1003

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java
@@ -988,7 +988,7 @@ public class ZigBeeDongleEzsp implements ZigBeeTransportTransmit, ZigBeeTranspor
 
     @Override
     public ZigBeeStatus setZigBeeChannel(ZigBeeChannel channel) {
-        if ((ZigBeeChannelMask.CHANNEL_MASK_2GHZ & channel.getMask()) == 0) {
+        if (channel != ZigBeeChannel.UNKNOWN && (ZigBeeChannelMask.CHANNEL_MASK_2GHZ & channel.getMask()) == 0) {
             logger.debug("Unable to set channel outside of 2.4GHz channels: {}", channel);
             return ZigBeeStatus.INVALID_ARGUMENTS;
         }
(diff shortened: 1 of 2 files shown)


cdjackson:
Thanks Stefan.

PR #1007, opened 2020-03-19
Fix TelegesisEnergyScanCommand response prompt in auto coder

### LINKED ISSUE #1006: TelegesisEnergyScanCommand response prompt is broken in auto coder
According to the Telegesis AT specification the prompt for an `TelegesisEnergyScanCommand` is `AT+ESCAN`. This is correctly implemented in the auto coder.

But the response to the command has the prompt `+ESCAN`, so the "AT" is missing.

I will create a PR for this.

EDIT: I have verified this with a telegesis stick and was wondering why the response wasn't matched correctly :)

triller-telekom:
Fixes #1006


cdjackson:
Thanks Stefan.

PR #1010, opened 2020-03-20
Telegesis scan for quietest channel

### LINKED ISSUE #1009: Telegesis scan for quietest channel
Related to #958 I like to propose an implementation for Telegesis similar to #1003 for Ember.

Passing the channel `ZigBeeChannel.UNKNOWN` to the `ZigbeeNetworkManager#setZigBeeChannel()` method now triggers a network scan and select the quietest channel.

Since Ember has already a resttriction in place for the 2GHz channels, see https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/73355bbc4563d0cee97c62c201dd9eaf77a23fd5/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java#L993-L1000

I am also doing a scan for the 2Ghz channels only, in the Telegesis implementation.

To be able to implement something like this, I need to get access to the private class `ScanResult` inside`TelegesisEnergyScanCommand`. I have changed the code generator to make those classes `public` in general, but I am not sure if this is a feasible solution.

Also I have fixed an issue in the autocoder that I introduced with #1007. because the "AT" was removed it crashed with an `StringIndexOutOfBoundEexception`.

triller-telekom:
Fixes #1009

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.telegesis.autocode/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/autocode/ClassGenerator.java
+++ b/com.zsmartsystems.zigbee.dongle.telegesis.autocode/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/autocode/ClassGenerator.java
@@ -44,6 +44,9 @@ public abstract class ClassGenerator {
     }
 
     private String toProperCase(String s) {
+        if (s.length() == 0) {
+            return s;
+        }
         return s.substring(0, 1).toUpperCase() + s.substring(1).toLowerCase();
     }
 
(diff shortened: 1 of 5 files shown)


cdjackson:
Do you need to run the autocoder as well? It seems you've made the change, but not updated the classes?

triller-telekom:
I did run the autocoder but it turned out only `TelegesisActiveScanCommand.java` and `TelegesisEnergyScanCommand.java` were affected by the internal class that I have made public now.

cdjackson:
Thanks Stefan.

PR #1014, opened 2020-03-26
Fix log message in ZclCluster

hsudbrock:
A tiny PR for a log message where the format parameters were positioned wrongly.

Before, the log message read like, e.g.: `0C4A/1: Error reading attribute Client in 2 cluster 25 - UNSUPPORTED_ATTRIBUTE`

Now, it reads like (for the above example): `0C4A/1: Error reading client attribute 2 in cluster 25 - UNSUPPORTED_ATTRIBUTE`


cdjackson:
Thanks Henning.

Issue #630, opened 2019-05-14
Remove deprecated cluster attribute read commands


cdjackson:
As per #591, in ```1.3.0``` we should remove the deprecated methods that were deprecated in ```1.2.0```.

tejas-n:
Can I pick this one up or is it planned to be done later? 😄
I'm assuming this has two parts?
1. Replacing the internal refrences to the deprecated methods and
2.  Modifying the auto encoder to not generate those methods and generate the new cluster classes?

cdjackson:
Hi Tejas,
I didn't want to do this too soon as I want to avoid breaking users systems :)

That said, if you wanted to pick up point 1 (replace internal references) then that would be great. I did do some, but I don't recall if I systematically went through and removed all internal uses or not. This could certainly be done now if you wanted, and would be appreciated.

The autocoder part is pretty simple since all methods are already marked deprecated, and it's just a case of removing them. I don't really want to do that for a while though as it will break systems that are still using these methods - my plan was to do this shortly before the update from V1.3.x to 1.4.0 which is probably a little way off at the moment.

Is there a reason you wanted to remove them? eg is it causing resource issues in your system?

tejas-n:
Thanks for the reply, Chris.
There's no particular reason I wanted to do it. It's just that I had a bit of free time on the weekend and I saw this issue which I assumed would be easy to do :)
I can certainly pick up point 1, I think I saw a few internal usages.

cdjackson:
Thanks Tejas. Picking up the first point is most of the work really and would be much appreciated. 

I guess your IDE will provide a list of all deprecated warnings.

Thanks.

PR #1017, opened 2020-03-28
Replace deprecated method calls

tejas-n:
* Remove references to specific attribute methods of the cluster classes which have been deprecated
* Prepares for #630 

Signed-off-by: Tejas Nandanikar <tejas3093@gmail.com>

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
@@ -397,9 +397,8 @@ public abstract class ZclCluster {
      * @return command future {@link CommandResult}
      */
     public Future<CommandResult> writeAttributes(List<WriteAttributeRecord> attributes) {
-        final WriteAttributesCommand command = new WriteAttributesCommand();
+        final WriteAttributesCommand command = new WriteAttributesCommand(attributes);
         command.setClusterId(clusterId);
-        command.setRecords(attributes);
         command.setDestinationAddress(zigbeeEndpoint.getEndpointAddress());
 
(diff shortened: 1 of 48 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/zcl/ZclCluster.java
```diff
@@ -1352,8 +1329,7 @@ private void handleDiscoverAttributes(DiscoverAttributesCommand command) {
             }
         }
 
-        DiscoverAttributesResponse response = new DiscoverAttributesResponse();
-        response.setAttributeInformation(attributeInformation);
+        DiscoverAttributesResponse response = new DiscoverAttributesResponse(true, attributeInformation);
```
cdjackson: This should not be ```true``` - I'm creating a PR to fix a bug here which we should fix in this PR as well.

com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/zcl/ZclResponseMatcherTest.java
```diff
@@ -61,7 +62,7 @@ public void testNonMatchingTransactionIds() {
     @Test
     public void testZdoResponse() {
         ZclCommand request = createOnCommand(new ZigBeeEndpointAddress(0,1), new ZigBeeEndpointAddress(1234,5), 5);
-        ZdoCommand response = new DeviceAnnounce();
+        ZdoCommand response = new DeviceAnnounce(123, new IeeeAddress("1234"), null);
```
cdjackson: Just for correctness, can you make the IeeeAddress argument 16 characters long so that it's a valid address.

tejas-n:
@cdjackson Thanks for the feedback, I've rebased and pushed the changes.

PR #1018, opened 2020-03-28
Set discoveryComplete flag when discovering attributes

cdjackson:
When requesting the list of supported attributes, the response was not properly handling the ```discoveryComplete``` flag which reports if there are more attributes than is being reported.

@tejas-n this will break your PR - please can you rebase #1017 against this once it's merged to master (I found this while reviewing your PR).


tejas-n:
@cdjackson Sure thing, thanks.

PR #1027, opened 2020-04-04
add ability to write byte array on serial port

mikomarrache:
The current ZigBeePort interface only provides the ability to write data one byte after the other.

For regular serial communication, that may not be a problem but for some other protocols, it causes a waste of bandwidth due to additional overhead for transferring every byte separately.

This pull request simply adds another method to the interface for writing a byte array instead. Of course, this is backward compatible change.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.cc2531/src/main/java/com/zsmartsystems/zigbee/dongle/cc2531/network/impl/CommandInterfaceImpl.java
+++ b/com.zsmartsystems.zigbee.dongle.cc2531/src/main/java/com/zsmartsystems/zigbee/dongle/cc2531/network/impl/CommandInterfaceImpl.java
@@ -296,9 +296,7 @@ public class CommandInterfaceImpl implements ZToolPacketHandler, CommandInterfac
     @Override
     public void sendRaw(int[] packet) throws IOException {
         synchronized (port) {
-            for (int i = 0; i < packet.length; i++) {
-                port.write(packet[i]);
-            }
+            port.write(packet);
         }
     }
 
--- a/com.zsmartsystems.zigbee.dongle.cc2531/src/test/java/com/zsmartsystems/zigbee/dongle/cc2531/Cc2351TestPacket.java
+++ b/com.zsmartsystems.zigbee.dongle.cc2531/src/test/java/com/zsmartsystems/zigbee/dongle/cc2531/Cc2351TestPacket.java
@@ -81,6 +81,10 @@ public class Cc2351TestPacket {
         }
 
         @Override
+        public void write(int[] bytes) {
+        }
+
+        @Override
         public int read(int timeout) {
             return read();
         }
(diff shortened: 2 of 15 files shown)


cdjackson:
Thanks. I'm certainly not against this and I agree that it makes sense. However you say this is backward compatible, but I don't think that is the case - this is a breaking change and definitely not backward compatible as implemented. 

Maybe you can make it backward compatible with a default implementation in the ```ZigBeePort``` interface but I think this would need to wait until shortly before the next major release (```1.4```) before it could be merged.

It seems the build is failing (although not all the CI tasks seem to be running for some reason)?

mikomarrache:
My bad - the change on the interface is indeed not backward compatible. I guess we will wait for the next major release then.

Regarding the build, it is fixed now. Some test classes I missed that were not implementing the new method.

cdjackson:
Thanks @mikomarrache. I think if we make this change there is no real point in keeping the old method since I don't think it's ever called is it?

mikomarrache:
The single byte method is still used. In any case, it's a convenient method for writing a single byte only which makes sense in some cases according to some places in the code.

mikomarrache:
Hi Chris,

I made the new method default on the interface. This way existing code should not be broken.

If all is good with the PR, would it be possible to merge it before 1.4.x?

Thanks

cdjackson:
> If all is good with the PR, would it be possible to merge it before 1.4.x?

I'll take another look closer to the time, but yes, let's target this for 1.4.  Thanks.

mikomarrache:
Is there an ETA for 1.4.0?

cdjackson:
@mikomarrache I will try and get 1.4 released in a month or so. I will likely add scenes and group support in January, and once that is all working we can look at getting 1.4 released.

Looking through this PR - I'd like to get this merged for 1.4 as we discussed earlier.  We provided the new method with a default implementation in the interface, but I'm wondering if we shouldn't change that.  By providing a default implementation we are going to move the point at which the developer discovers the system no longer works from compile time to run time. Either way, it won't work, and by using a default implementation we're kind of masking that - we might as well make it obvious right from the start.

What do you think?

(Merry Christmas :) and thanks for all your inputs over the past year - much appreciated).

mikomarrache:
I changed the method to default thinking that it would be possible to merge the PR on 1.3.x. Because it will be released with 1.4, and that it is agreed this version introduces breaking changes, I think it is best to declare the new method as non default to force client code to implement it. I will revert the last commit then.

Merry Christmas :)

cdjackson:
>  I think it is best to declare the new method as non default to force client code to implement it. I will revert the last commit then.

@mikomarrache I fully agree with your point above so thanks for reverting this.

Now that 1.4.0-SNAPSHOT is the current build, I'm happy to merge this if you're happy with this? Sorry it's taken a while to get to this point.

mikomarrache:
Yes please - you can merge. Thanks

Le dim. 28 mars 2021 à 05:36, Chris Jackson ***@***.***> a
écrit :

cdjackson:
@mikomarrache there seems to be a couple of issues with this PR - can you take a look please. One is simple - please run the `license:format` goal to update the year. Even with this though it fails here and I think I got Travis working yesterday and it's also showing a failure...

mikomarrache:
@cdjackson are there still issues with this PR?

cdjackson:
Let me check tonight - I have to go out soon...

PR #1037, opened 2020-04-14
Fix autocoder array count generator

### LINKED ISSUE #1035: Overlapping attribute ids
Some cluster attributes ids are overlapping, see for example in ZclMeteringCluster.java

```
public static final int ATTR_PREVIOUSDAY16CONSUMPTIONRECEIVED = 0x0431;

public static final int ATTR_CURRENTWEEKCONSUMPTIONRECEIVED = 0x0431;
```
I believe the issue is here:
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/759570d1975c25395c87c41507046e2340b608a2/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeZclClusterGenerator.java#L176-L177

I think the id shouldn't be generated as (attribute.code + arrayCount) but as (attribute.code + arrayCount - attribute.arrayStart)

cdjackson:
Fixes #1035 

Thanks for reporting this @spudwebb - you probably should check this, and more to the point, you should check your implementation as from the quick look I had I suspect it's wrong as the enum/const looks to be wrong.

Eg -:

image

My fix is -:
image

Your is -:
image


spudwebb:
you're right, thanks for the heads up!

PR #1041, opened 2020-04-29
ezsp - add support for custom frames

mikomarrache:
I added 2 entries to the file ezsp_protocol.xml and run the autocoder.

One entry is a command that can be used to send custom frames to the NCP via EZSP.
The other entry is an handler for receiving custom messages from the NCP.

PR #1042, opened 2020-04-30
fix compilation issue introduced in PR #1041

mikomarrache:


cdjackson:
Have you had to manually make these changes? If so we probably should address the autocoder to fix this otherwise these changes will be automatically reverted.

mikomarrache:
Ok - I think I got it. I missed the auto_size element.

Is that ok?

cdjackson:
There is something still failing with this - I'll merge this since it looks ok and will sort out any subsequent issues.

cdjackson:
Not sure why the CI seemed to fail - it's compiling fine locally here.

mikomarrache:
I guess the CI failed because this PR doesn't compile on its own. It depends on the other one (#1041).

cdjackson:
But wasn’t that already merged? I’m pretty sure I merged, that, then you provided the fix for the compilation problem…

Anyway, it seems to be compiling now, and certainly worked ok locally here.

Thanks.

PR #1043, opened 2020-04-30
Add Ember NCP commands

mikomarrache:
Add 2 new commands:

- Ability to get XNCP information from NCP
- Ability to send custom frames


cdjackson:
Thanks for this. My only question is if we need to add the extra class to return the information? My preference is to use the same concept that the other methods use - ie to return the `EzspXxxResponse` frame. I would prefer to keep things standard, and I think introducing extra classes to return data would introduce a lot of extra classes for no real benefit.

What do you think @mikomarrache?

mikomarrache:
Agreed. I pushed a new change.

cdjackson:
Thanks @mikomarrache

PR #1045, opened 2020-05-04
Return error when Ember network not found

### LINKED ISSUE #1044: ZigBeeDongleEzsp should do formNetwork() if not joined
I just stumbled across line 478 in this code:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/5a91440b8e4774b9e81bd58e57a1da7cfbd87713/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java#L476-L502

Here we detect that the chip has not yet joined a network. However, we do line 496, i.e. the `formNetwork` call only if `reinitialize` has been set to `true`.

So I am wondering whether we should also set it to `true` if the chip reports us in line 478 that the chip isn't joined.

Is this a bug or is there a reason to only call `formNetwork` of the `reinitialize` flag has been set from outside?

cdjackson:
@triller-telekom please let me know if this is ok for you.

Fixes #1044


triller-telekom:
Thanks Chris, LGTM.

PR #1048, opened 2020-05-05
Fix ZigBeeDongleEzsp startup

### LINKED ISSUE #1047: Bug in ZigBeeDongleEzsp.startup introduced this morning
Unfortunately the PR https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1045/

has introduced a bug. Now it is not possible at all to start a new network on a brand new factory reseted chip :(

Even if `reinitialise` is set to `true`, we will get a `EmberStatus.EMBER_NOT_JOINED` and thus return without forming a new network :(

So we should only return in the case `reinitialise` is `false` and we get `EmberStatus.EMBER_NOT_JOINED`.

triller-telekom:
Fixes #1047


cdjackson:
Ooops - thanks.

PR #1050, opened 2020-05-05
Network address is not set during initialization

mikomarrache:
Therefore, the error should not be logged during initialization.


mikomarrache:
Finally, I found no reason to add the local node during initialization, so I remove the addLocalNode() call.

Moreover, during startup, the code expected a local node to be present, which was not the case. Therefore, I also removed this check.

Note that the local node is added later when the network is set to ONLINE.

cdjackson:
I'm a little hesitant to merge this due to the following change -:

This PR will mean that the local node is now only **GUARANTEED** to be available after the network is online. Previously it would be **GUARANTEED** to be available earlier, although the network address would not necessarily be correct until the local node is ONLINE.

The local node **MAY** be available earlier if the network is restarting as it will be restored from the data store but users can no longer rely on this.

---

Probably this is ok, but I'll leave this a few days at least to allow anyone to comment.

cdjackson:
This works ok here on the Ember at least so I'll merge this.

PR #1058, opened 2020-05-15
ZigBeeNodeServiceDiscoverer should use list instead of set for its tasks

### LINKED ISSUE #1057: ZigBeeNodeServiceDiscoverer has random order of tasks
The `ZigBeeNodeServiceDiscoverer` uses an `HashSet` for its `discoveryTasks`.

Sometimes we see that we cannot discover devices and while scanning through all the logs over the past month I found a pattern. It failed for sure if the order of discovery tasks was wrong.

In the cases where it failed the NWK_ADDRESS task was the last in the set, see for example here:

```
2020-05-14T09:18:58.218+0200   463  DEBUG          [symbolicName=com.zsmartsystems.zigbee, className=com.zsmartsystems.zigbee.app.discovery.ZigBeeNodeServiceDiscoverer] - 00124B0007515258: Node SVC Discovery: starting new tasks [POWER_DESCRIPTOR, NODE_DESCRIPTOR, ACTIVE_ENDPOINTS, NWK_ADDRESS]
```

But the other tasks need the network address in order to function.

I have attached an example log for the message above where one can see that the first 3 tasks fail after 12 retries each and the last one is successful. But of course this means that the discovery has failed.

[zsmart.txt](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/files/4634340/zsmart.txt)

triller-telekom:
Fixes #1057

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
@@ -197,7 +197,7 @@ public class ZigBeeNodeServiceDiscoverer {
      *
      * @param newTasks a set of {@link NodeDiscoveryTask}s to be performed
      */
-    private void startDiscovery(Set<NodeDiscoveryTask> newTasks) {
+    private void startDiscovery(List<NodeDiscoveryTask> newTasks) {
         // Tasks are managed in a queue. The worker thread will only remove the task from the queue once the task is
         // complete. When no tasks are left in the queue, the worker thread will exit.
         synchronized (discoveryTasks) {
@@ -676,7 +676,7 @@ public class ZigBeeNodeServiceDiscoverer {
      */
     public void startDiscovery() {
         logger.debug("{}: Node SVC Discovery: start discovery", node.getIeeeAddress());
-        Set<NodeDiscoveryTask> tasks = new HashSet<NodeDiscoveryTask>();
+        List<NodeDiscoveryTask> tasks = new ArrayList<NodeDiscoveryTask>();
 
         // Always request the network address unless this is our local node - in case it's changed
         if (!networkManager.getLocalNwkAddress().equals(node.getNetworkAddress())) {
@@ -710,7 +710,7 @@ public class ZigBeeNodeServiceDiscoverer {
 
         } else {
             logger.debug("{}: Node SVC Discovery: Update mesh", node.getIeeeAddress());
-            startDiscovery(new HashSet<>(meshUpdateTasks));
+            startDiscovery(new ArrayList<>(meshUpdateTasks));
         }
     }
 


cdjackson:
Thanks Stefan,
Does this really solve the issue you're reporting? Changing to a list will retain insertion order, so probably it is now better. However, I'm not sure if we get the situation that tasks can be added out of order - if I remember correctly the class adds tasks on top of those that may already be there, so it's possible that tasks exist when the "get nwk address" task is added.

I've not checked this for sure though, but it's worth checking. I just wonder if we want to ensure order, if we should use some sort of sorted list, or do a sort before kicking off tasks?

hsudbrock:
I did a check on the issue that you, @cdjackson, mention with respect to adding discovery tasks on top of other tasks that may already be there.

TLDR: This should not be an issue (but it's rather implicit, and one could think of improving this).

There are two situations in which discovery tasks are added to the queue:
1. When starting the discovery for a node in [`ZigBeeNodeServiceDiscoverer.startDiscovery()`](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/531addfbab8732563493ee14d2519bc986e91f4a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java#L677), which adds (a subset of) the discovery tasks `NWK_ADDRESS`, `NODE_DESCRIPTOR`, `POWER_DESCRIPTOR`, and `ACTIVE_ENDPOINTS`.
2. When updating the mesh in [`ZigBeeNodeServiceDiscoverer.udpateMesh()`](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/531addfbab8732563493ee14d2519bc986e91f4a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java#L706), which adds the discovery tasks `NEIGHBORS` and `ROUTES`.

I think that the `ZigBeeDiscoveryExtension` currently ensures that `startDiscovery()` is always called before `meshUpdate()` (because `meshUpdate` is only called on nodes in the map `ZigBeeDiscoveryExtension.nodeDiscovery`, and `startDiscovery()` is called immediatey after adding a node to that `nodeDiscovery` map, and synchronization ensures that there cannot be a race in which `meshUpdate()` is called between adding a node to that map and starting discovery for that node). 

In consequence, it is ensured (with Stefan's change from `Set` to `List`) that the `NWK_ADDRESS` discovery task is always the first to run. But I think that this is rather implicit currently. To make this more explicit, one could, for instance, always add a `NWK_ADDRESS` discovery task in `ZigBeeNodeServiceDiscoverer.startDiscovery(List<...> newTasks)` if `newTasks` does not contain that task and the network address is not known for the node. Do you think this would be useful?

cdjackson:
So let's continue to debate this `NWK_ADDRESS` issue here @hsudbrock 

Thinking about this some more - firstly, I tend to agree with you - we must be sure we know the network address, or things won't work, and it would be better to explicitly check this. The question then becomes - how do we know that we know the network address? (ie what check do we use).

We could use `== null` - that's the simplest and most obvious. If it's `null`, then we don't have a network address and we absolutely need to have this. Simple ;)

But, thinking about this some more - maybe that's not enough. The network address can change if an ED moves parent. In theory we should get notified and the `NWK_ADDRESS` gets updated in the `ZigBeeNode`. If we miss this, or it happens when the binding is down, then we ALWAYS request the `NWK_ADDRESS` on startup. So it should be fine - right?

Based on this, all we need is the `null` check.

But I'm trying to think if we ever perform a network address rediscovery other than at startup. We could just do it as a matter of course with every mesh update - to be 100% sure that we know the **current** network address. Or (or possible AND) we could do this if we start seeing requests to a device fail.

The second part of the question is not related to this issue so I might raise this as a separate issue, but the question here is -:

Do we simply ALWAYS request the `NWK_ADDRESS` as part of a mesh update? The only downside is a bit more traffic - if we work on the assumption that the mesh update is not happening super often, does that matter?

triller-telekom:
> Do we simply ALWAYS request the NWK_ADDRESS as part of a mesh update? The only downside is a bit more traffic - if we work on the assumption that the mesh update is not happening super often, does that matter?

I think this should be feasible to do. I mean those IEEE address requests aren't super large and if there is traffic because of the mesh update anyway, I think these few more bytes won't hurt.

cdjackson:
Thanks Stefan - This is also my view. I'll update the PR to do two things -:

1. I will add the null check as mentioned above so that we guarantee that we have at least an address.
2. I will add `NWK_ADDRESS` to the list of actions to perform as part of the mesh update - this means that users can decide if they want to remove it from the regular requests.

I think this is a good middle ground that ensures we cover cases where we don't know the address, but gives users the option if they want to have the extra network traffic or not, and in all cases this should now always be the first action to run by using the `PriorityQueue`.

cdjackson:
Replaced by #1064

PR #1060, opened 2020-05-18
Remove the 'desiredVersion' parameter from the Ember NCP getVersion command

### LINKED ISSUE #1059: Desired version parameter is ignored in EmberNcp.getVersion()
The method [`EmberNcp.getVersion()`](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/80f99c1466158a552d2c39dd1be4abd5e716172e/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/EmberNcp.java#L181) allows to provide the 'desired version' as a parameter. However, this parameter is ignored, and the desired version is always set to the static `EzspFrame.ezspVersion` (currently `4`).

I guess the `desiredVersion` parameter can be removed from `EmberNcp.getVersion` (or do you think, Chris, that this method should rather be rewritten to really allow for setting a desired version on the chip?).

hsudbrock:
Resolves #1059

As written in #1059, another option would be to actually use the `desiredVersion` parameter in the method (instead of using the static version value from `EzspFrame`), but as the implementation of `ZigBeeDongleEzsp` adapts the static version value in `EzspFrame` in case the chip supports another version only it sounded more sensible to me to remove the parameter.

If you disagree, Chris, I can adapt the PR.

### DIFF
--- a/com.zsmartsystems.zigbee.console.ember/src/main/java/com/zsmartsystems/zigbee/console/ember/EmberConsoleNcpStateCommand.java
+++ b/com.zsmartsystems.zigbee.console.ember/src/main/java/com/zsmartsystems/zigbee/console/ember/EmberConsoleNcpStateCommand.java
@@ -94,7 +94,7 @@ public class EmberConsoleNcpStateCommand extends EmberConsoleAbstractCommand {
         String mfgName = ncp.getMfgName();
         String mfgBoard = ncp.getMfgBoardName();
         int customVersion = ncp.getMfgCustomVersion();
-        EzspVersionResponse version = ncp.getVersion(5);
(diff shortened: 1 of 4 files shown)


hsudbrock:
Shall this (new?) `codecov/patch` check enforce that unit tests are added when changing functionality that has no unit tests yet?

cdjackson:
Thanks Henning. This looks fine to me.

Regarding the codecov checks - this isn't new if that's what you mean. It is meant to at least indicate if there is coverage, and therefore tests, for new code, but depending on the exact code if it's "enforced".

Some CI tasks seem to not work occasionally at the moment (eg Travis!) so I need to look at that.

PR #1061, opened 2020-05-18
Timeout sending EZSP transactions after 10 seconds

### LINKED ISSUE #1056: ASH/EZSP initialization thread hangs forever if Ash reset frame is not acknowledged
In some error conditions, the `AshFrameRst` frame is not acknowledged with an `AshFrameRstAck` frame. (This happens, e.g., when the baud rate for the serial connection is wrongly configured.)

Currently, if this happens when connecting to the frame handler [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/80f99c1466158a552d2c39dd1be4abd5e716172e/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java#L1242) during the `ZigBeeDongleEzsp` initialization, the initialization hangs indefinitely in the `ncp.getVersion(..)` call [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/80f99c1466158a552d2c39dd1be4abd5e716172e/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java#L1248).

It hangs indefinitely because `ncp.getVersion()` leads to calling `AshFrameHandler.sendEzspTransaction`, which waits for the EZSP transaction to finish [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/80f99c1466158a552d2c39dd1be4abd5e716172e/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java#L786), but the transaction won't finish before the `AshFrameHandler` has successfully connected (the frame handler sends out `AshFrameRst` frames again and again, but they get never acknowledged in the error condition mentioned above).

I would like to handle this more gracefully, so that the thread trying to initialize the EZSP protocol does not hang forever.

My suggestion is to fail the initialization after a timeout (say, 10 seconds?). I see two options for this:

1. Wait until the frame handler has connected before issuing the `getVersion` command (by inspecting the currently private variable `AshFrameHandler.stateConnected`).
2. Add a timeout when resolving the future in `AshFrameHandler.sendEzspTransaction` [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/80f99c1466158a552d2c39dd1be4abd5e716172e/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java#L786).

I would prefer variant 2, because it avoids the situation that we check that the frame handler is connected, but it disconnects for whatever reason before we can send out the getversion command.

I could provide a PR, but I think it would be good to align here first on how to improve this. Any opinions?

hsudbrock:
Resolves #1056 

I have added the timeout for both ASH and SPI frame handlers.

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java
@@ -93,6 +93,11 @@ public class AshFrameHandler implements EzspProtocolHandler {
 
     private final static int ASH_MAX_LENGTH = 220;
 
+    /**
+     * Timeout after which sending an EZSP transaction is aborted.
+     */
+    private static final long EZSP_TRANSACTION_TIMEOUT_SECONDS = 10;
+
     private Integer ackNum = 0;
     private int frmNum = 0;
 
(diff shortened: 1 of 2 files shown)


cdjackson:
Thanks Henning.

PR #1063, opened 2020-05-22
Fix incorrect CommandResult returned for a command sent with disableDefaultResponse set to true

### LINKED ISSUE #1062: Issue with the CommandResult returned for a command sent with disableDefaultResponse set to true
**Describe the bug**
If a command is sent with `disableDefaultResponse` set to true, the returned `CommandResult` yields the following results even if the command was sent successfully

`commandResult.isSuccess()` returns false
`commandResult.isError()` returns true
`commandResult.isTimeout()` returns true

**Expected behavior**
On a successful transmission:
`commandResult.isSuccess()` should returns true
`commandResult.isError()` should return false
`commandResult.isTimeout()` should return false

**Source of the issue**
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/531addfbab8732563493ee14d2519bc986e91f4a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransaction.java#L479-L489
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/531addfbab8732563493ee14d2519bc986e91f4a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransaction.java#L402-L419

From the above code it looks like if `RX_ACK` is received and the `disableDefaultResponse` was set, it means that the system is not expecting a default response which is why it completes the transaction. But when calling `completeTransaction(completionCommand)` it passes null as parameter since it wouldn't have received any command. And looking at `CommandResult`, the `isTimeout()` and `isError()` will always return true if the completionCommand is null.

**Proposed Solution**
1. Pass another parameter to `CommandResult` constructor like `commandSentSuccessfully` along with the `completionCommand` so that you can have a command which completes successfully but doesn't expect any response(like the ones which don't expect `DefaultResponse`
2. Deprecate the constructor which expects just the `completionCommand`

I'll create a pull request for reference. Can you let me know your thoughts, please? Cheers!

tejas-n:
Fixes #1062 
Signed-off-by: Tejas Nandanikar <tejas3093@gmail.com>

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/CommandResult.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/CommandResult.java
@@ -23,20 +23,44 @@ public class CommandResult {
      * The response command.
      */
     private final ZigBeeCommand response;
+    private final ZigBeeStatus commandSentSuccessfully;
 
     /**
      * Constructor which sets the received response command or null if timeout occurs.
      *
      * @param response the response command.
+     * @deprecated Use {@link #CommandResult(ZigBeeStatus, ZigBeeCommand)} instead
      */
+    @Deprecated
     public CommandResult(final ZigBeeCommand response) {
+        if (response != null) {
+            this.commandSentSuccessfully = ZigBeeStatus.SUCCESS;
+        } else {
+            this.commandSentSuccessfully = ZigBeeStatus.FAILURE;
+        }
+        this.response = response;
+    }
+
+    /**
+     * Constructor which sets the received response command and if the command was sent successfully.
+     *
+     * @param commandSentSuccessfully if the command was sent successfully
+     * @param response                the response command
+     */
+
+    public CommandResult(final ZigBeeStatus commandSentSuccessfully, final ZigBeeCommand response) {
+        this.commandSentSuccessfully = commandSentSuccessfully;
         this.response = response;
     }
 
     /**
      * Constructor for timeout situations.
+     *
+     * @deprecated Use {@link #CommandResult(ZigBeeStatus, ZigBeeCommand)} instead
      */
+    @Deprecated
     public CommandResult() {
+        commandSentSuccessfully = ZigBeeStatus.FAILURE;
         response = null;
     }
 
@@ -55,7 +79,7 @@ public class CommandResult {
      * @return TRUE if timeout occurred
      */
     public boolean isTimeout() {
-        return response == null;
+        return commandSentSuccessfully != ZigBeeStatus.SUCCESS;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


cdjackson:
Thanks @tejas-n .

I suspect I'm missing something here - the code looks fine in general, but what I don't quite understand is the use. In this change, the only place that `CommandResult` is instantiated, `CommandSentSuccessfully` is set to `true`.

If it is always `true` then I'm not sure I understand the need, so I guess I miss something?

tejas-n:
Thanks for the reply, @cdjackson. Calling the default constructor actually sets it to false and the place where the default constructor is called is here:
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/531addfbab8732563493ee14d2519bc986e91f4a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionFuture.java#L64-L85

Did you get a chance to look at the linked issue? I guess `CommandResult` needs something which would imply the command was sent successfully even if the library didn't receive any response(like in the case of commands not expecting even `DefaultResponse`)?

cdjackson:
Thanks. But my point is that if this constructor is always called with `true`, then why is it needed?  I think I'm still missing something, but you always seem to call this constructor with `commandSentSuccessfully` set to `true` so why have this parameter? Why not just stick with the constructor you've deprecated?

tejas-n:
But with `public CommandResult(final ZigBeeCommand response)` constructor, we don't have a way to specify that a command which was sent with `disableDefaultResponse` set to true was successful(since such commands won't receive any DefaultResponse) even if APS ACK was received. Am I correct?

Maybe we also deprecate the default `CommandResult` constructor and change `ZigBeeTransactionFuture` to use:
```
set(new CommandResult(false, null)
```
What do you think?

cdjackson:
From what I see in this PR, there are two ways to instantiate `CommandResult` -:

`CommandResult()` and `CommandResult(true, response)`.

Given that the `CommandResult(completed)` is ALWAYS `true`, I can't see the point in this parameter. I guess I'm still missing something somewhere - when is this called with `false` (other than in the tests)? If it's not used, or is always a static value, then I'm having trouble understanding it's purpose. I can understand why you want to add it, but it just doesn't seem to be used (does it?) - it's **always** `true`.

cdjackson:
Or maybe there are other calls to `CommandResult(response)` that you've not migrated? If so, and you are deprecating this constructor, please ensure that the deprecated constructor is not used within the framework.

tejas-n:
Okay, I guess I get what you mean. I've updated the PR, can you review it, please?
1. I've deprecated the default `CommandResult` constructor as well.
2. Updated the library so that only the new constructor is used.

cdjackson:
Thanks @tejas-n. Can I make one last suggestion. Instead of using `boolean` can we instead use `ZigBeeStatus`. This just makes it more readable, and expandable in future if we wanted to pass some other status into the constructor.

So we have `ZigBeeStatus.SUCCESS`, `ZigBeeStatus.FAILURE` for now.

tejas-n:
Thanks, @cdjackson, that makes sense. I've updated the PR and also rebased it onto the latest master commit.

cdjackson:
Thanks @tejas-n

PR #1064, opened 2020-05-26
Ensure discovery tasks are run in order

### LINKED ISSUE #1057: ZigBeeNodeServiceDiscoverer has random order of tasks

cdjackson:
This is an alternative suggestion to resolve #1057 and replaces #1058.

@hsudbrock I think this is simpler than your or @triller-telekom suggestion and I think it solves the issue. This simply changes the  `discoveryTasks `  implementation of `Queue` from `LinkedList` to `PriorityQueue` - thus ensuring the discovery order is based on the `enum` ordinal.

I think this will resolve the issue Stefan raised. I also think this is safer - we guarantee the order of discovery as there are potentially other dependencies - eg we can't request the endpoint simple descriptor until we have the node descriptor.

What do you think?

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeDiscoveryExtension.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeDiscoveryExtension.java
@@ -74,10 +74,13 @@ public class ZigBeeDiscoveryExtension
     private boolean extensionStarted = false;
 
     /**
-     * List of tasks to be completed during a mesh update
+     * List of tasks to be completed during a mesh update.
+     * We want to get the neighbors and routes so we have visibility of the mesh. We also default to requesting the
+     * network address to ensure that it hasn't changed.
      */
     private List<NodeDiscoveryTask> meshUpdateTasks = Arrays
-            .asList(new NodeDiscoveryTask[] { NodeDiscoveryTask.NEIGHBORS, NodeDiscoveryTask.ROUTES });
+            .asList(new NodeDiscoveryTask[] { NodeDiscoveryTask.NWK_ADDRESS, NodeDiscoveryTask.NEIGHBORS,
+                    NodeDiscoveryTask.ROUTES });
 
     @Override
     public ZigBeeStatus extensionInitialize(ZigBeeNetworkManager networkManager) {
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNodeServiceDiscoverer.java
@@ -12,8 +12,8 @@ import java.util.Calendar;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.LinkedList;
 import java.util.List;
+import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.Random;
 import java.util.Set;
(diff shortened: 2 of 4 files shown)


hsudbrock:
Thanks for this alternative suggestion, Chris. I agree that using a queue that automatically ensures that the discovery tasks are executed in the correct order is safer.

What I think would be nice is to make the dependencies between the discovery tasks more explicit; this could be done by explicitly defining an ordering on the enum (instead of using the default ordering on the enum); nevertheless, probably a comment on the enum with the discovery tasks stating that the NWK_ADDRESS must be run before all other tasks and that the node descriptor task must be run before the simple descriptor task would already suffice. Wdyt?

Side note regarding my comment on Stefan's pull request: I think that using the sorted queue does not address the issue that I raised in that comment, i.e., that it is not enforced that all needed tasks are always run. It is, e.g., still possible to enqueue the mesh update tasks without also enqueuing the task for retrieving the network address. However, as this is currently ensured implicitly by the way in which the `ZigBeeDiscoveryExtension` uses the `ZigBeeNodeServiceDiscoverer`, this is nothing that must be added to have the discovery function correctly.

cdjackson:
To summarise the discussion elsewhere, to ensure that the network address is known, the following checks / changes are made -:

1. I will add the null check as on the node network address so that we guarantee that we have at least an address.
2. I will add `NWK_ADDRESS` to the list of actions to perform as part of the mesh update - this means that users can decide if they want to remove it from the regular requests. It is considered useful to do this to ensure that we capture any network address changes.

@hsudbrock regarding your point -:

> What I think would be nice is to make the dependencies between the discovery tasks more explicit; this could be done by explicitly defining an ordering on the enum

I don't see a lot of point in adding an order constructor versus using the ordinal. We could make things even more complex and have an enum constructor that takes a list of dependencies so that it auto prioritises to ensure dependencies are met, but that sounds overly complex. If you're happy to use the ordinals, then I propose that we go with that?

triller-telekom:
Thanks Chris. This PR will solve my issue https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/1057

and make https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1058 obsolete.

Code-wise it looks good to me!

hsudbrock:
Chris, I agree that adding code to explicitly define the ordering for the enum is too much for the gain; that is why I suggested to simply add a comment that the order is significant and why that is.

I would suggest to add a comment to the enum similar to the following:

` Please note that the ordering of the tasks in the enum is significant; a task that needs to be run before another task must be placed before that other task in the enum. Examples are that the network address must be discovered before all other tasks can be run, and that the node descriptor must be discovered before the endpoint simple descriptor.`

cdjackson:
Thanks Henning - I'll update as suggested.

Issue #1096, opened 2020-07-26
Reduction of broadcast traffic


cdjackson:
During some recent testing with a reasonably large network, it has become obvious that the framework is sending too many broadcasts - especially on startup.  On the Ember NCP this causes a lot of failures - while the framework paces the number of broadcasts being sent, it does not have visibility of all other broadcasts being sent by the coordinator itself - eg APS broadcasts, APS multicasts, route discoveries, ZDO announcements or broadcast requests, PAN ID / channel / NWK key update notifications. The ZigBee (or 802.15.4) spec limits the number of broadcasts that can be sent to around 8 in an 8 second period - the transaction manager respects this (currently only releasing a broadcast every 1.2 seconds) but if there are other broadcasts going on, as there will be when the network is starting and routes are being discovered, then many of these will fail.

So, looking at ways to reduce this...

Currently, the system always attempts to get the network address of a device by sending a broadcast on startup - the assumption being that it _might_ have changed. However, routers should not change address, and ZEDs may change address if they change parent, but that does not happen often. Therefore, it is propose to always start this discovery with a unicast, and only use broadcast as a last resort.

The `ZigBeeNetworkDiscoverer` also uses broadcasts for a similar reason - this was actually written before the transaction manager was implemented, and was not updated. The transaction manager performs retries of frames that time out, so the retires in this class should be removed.

cdjackson:
Closed by #1099

PR #1099, opened 2020-08-02
Reduce broadcasts by using unicast where possible

cdjackson:
This reduces the reliance on broadcasts during discovery. The framework will now attempt to use unicast messages during discovery - if that fails, it may drop back to broadcast.

Too many broadcast requests during the initialisation of an application can seriously hinder the performance. Only a small number of broadcasts are permitted to be sent in a network - approximately 1 per second. This must include all broadcasts from all devices, and this constraint currently causes a number of errors sending broadcasts in large networks.

This PR also removes the retry mechanism that was in the discovery service since this is no longer required since the transaction manager was added.


cdjackson:
Closes #1096

PR #1103, opened 2020-08-11
Fix ZigBeeNetworkDiscoverer always using broadcast

### LINKED ISSUE #1102: ZigBeeNetworkDiscoverer is still using broadcast, for network address queries
I think I found a bug in:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/5c62379d7b6bb4fb58a6c392ed9bf19425140bad/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java#L204

Shouldn't it rather be ´if (ieeeAddressResponse == null) { ` there?

Given the comment
> Starts a discovery on a node. This will send a {@link NetworkAddressRequest} as a unicast, and if no response is the response to trigger a full discovery.

in `ZigBeeDiscoveryExtension.java` we should only do a broadcast if the unicast was NOT successful, i.e. the response was `null`.

triller-telekom:
Fixes #1102


cdjackson:
Thanks @triller-telekom

PR #1105, opened 2020-08-12
Reduce Ember NCP max hops to 8

cdjackson:
This reduces the default number of hops in an effort to improve network performance.
In the Ember NCP using a large number of hops will increase route table size. Additionally, if there are too many hops the system is likely to hit other issues such as timeouts due to the delays through the network.

10 is the maximum recommended value in discussion with a Silabs FAE. Setting this to 8.


triller-telekom:
> 10 is the maximum recommended value in discussion with a Silabs FAE

And 10 is a rather large number I would say. At least for "Smarthome" home installations. To get 30 hops that we currently set, one has to buy 30 devices capable of being used as a repeater first ;-)

cdjackson:
Yes, 10 is still quite large, but a lot less than 30 (I think it's 20 less :) ).

We could reduce further (6 or 8), but of course these settings can all be adjusted by implementations without changing the framework. 10 seems a reasonable maximum for even the largest environment, but it might be that during my current work I might find a slightly lower number is better.

PR #1107, opened 2020-08-14
Fix using always using broadcast in 2nd discovery

### LINKED ISSUE #1106: ZigBeeNetworkDiscoverer has two discovery methods that need to be fixed
I thought #1103 had fixed the bug with the discovery always using the broadcast address, however there is a SECOND method that suffers from the same bug.

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/master/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java#L268

It took me a while to find this as the methods look similar, and the difference is the `IeeeAddressRequest` vs `NetworkAddressRequest` which are created.

Also one of them does not implement the reaction on an interrupt:
```
if (Thread.currentThread().isInterrupted()) {
                        return;
                    }
```

I will provide a fix to get them back in sync, but we should certainly refactor this at some point, because code duplication is always bad...

triller-telekom:
Fixes #1106 and some test cases to make the build green again


cdjackson:
Thanks @triller-telekom

PR #1110, opened 2020-08-18
Increase the default timeout for transaction futures

### LINKED ISSUE #1109: Transactions should start their timeout timer only if they are send and not just enqueued
From the discussion of the closed ticket https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/1104#issuecomment-674538471 I think that this is a bug too:

> The only real problem I see here is with the transaction timeout. This timeout should not actually be in force, so there is a bug here I think. The issue is there is this default 12 second timeout in the get waiting for the future - this really should not be there IMHO. It causes the command to fail even though it is still waiting in the queue to be sent.

I am just wondering where to start fixing this bug? I think the goal is to only start the timeout timer once we have tried to send out the transaction at least once. Before that, a timeout should not occur, since the transaction is still enqueued and will eventually be send out.

In `ZigbeeTransaction` I see 2 times outs already:

```
/**
     * The amount of time (in milliseconds) from when the command is sent to the transport, until when the transport
     * acknowledges it has been transmitted.
     * This must account for the total time from when the data is sent, until the response is expected, included any
     * queueing delay in the transport.
     */
    private final static int TRANSACTION_TIMER_1 = 10000;
    private int timeout1 = TRANSACTION_TIMER_1;
```

and

```
/**
     * The amount of time (in milliseconds) to wait for a response from the transport once the command has been
     * transmitted.
     */
    private final static int TRANSACTION_TIMER_2 = 8000;
    private int timeout2 = TRANSACTION_TIMER_2;
```

I think the first one is just what we need. However in #1104 the `ZigBeeTransation.getFuture().get()` returns too fast.

That's why I think the bug must be in here:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/878754782735f559d7d55a93127de21311e4a460/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionFuture.java#L74-L86

So I am wondering whether the timeout in `get` somehow has to consider the `TRANSACTION_TIMER_1` from `ZigbeeTransaction`.

Any suggestions?

triller-telekom:
Fixes #1109

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionFuture.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/transaction/ZigBeeTransactionFuture.java
@@ -28,8 +28,12 @@ public class ZigBeeTransactionFuture implements Future<CommandResult> {
 
     private boolean cancelled = false;
 
+    /**
+     * We set the timeout to 5 minutes to be long enough to allow the transaction manager to complete its queuing and
+     * eventual retries
+     */
     // Not final for tests
-    private static long TIMEOUT_MILLISECONDS = 12000;
+    private static long TIMEOUT_MINUTES = 5;
 
     /**
      * Sets the {@link CommandResult}.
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks Stefan,
Can I suggest adding some javadoc (or comment at least) above this definition to explain briefly why it is 5 minutes? eg it has to be long enough to allow the transaction manager to complete all queuing and retries.....

Just so we don't forget in a years time and change it back :)

triller-telekom:
@cdjackson Thanks for the feedback. I have just seen it as my github notifications sometimes get stuck in gmail...

I have amended by commit and added a comment. Please have a look again.

cdjackson:
Thanks Stefan.

PR #1112, opened 2020-08-18
Fix ZigbBeeTransactionTutureTest

triller-telekom:
Somehow my IDE is configured wrong, so the tests were not recompiled after my last change.


triller-telekom:
Fixes the build error introduced by #1110

cdjackson:
Thanks.

PR #1163, opened 2020-11-14
Fix notification when device left

### LINKED ISSUE #1161: Node listeners not called when node becomes OFFLINE
I think there may be a problem with the following code.
In this case updateNode does nothing because we get the node from the network manager list and we set its state to OFFLINE, so that when ZigBeeNode.updateNode is called it will not detect any change because it compares the node instance with itself.

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/e1dd507e8faf77ab7f37e9ca03da95f93139d7a2/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java#L1187-L1200

cdjackson:
Fixes #1161 


cdjackson:
@spudwebb please let me know if this fixes your issue.

spudwebb:
I found this issue in https://github.com/Mr-Markus/ZigbeeNet and I checked this project source code to see if there was the same problem.  So, I can't really test your fix, but it looks good. I made a similar one in ZigbeeNet and it works there.

cdjackson:
Thanks @spudwebb - understood and appreciated.

PR #1166, opened 2020-11-19
Add console command to request neighbors

mikomarrache:


cdjackson:
Thanks @mikomarrache. Looks good to me. We need to consolidate this function somewhere as it's now repeated in a number of places (well - it was already - this just adds one more :) ).

PR #1168, opened 2020-11-19
Add console command to read all attributes of a cluster

mikomarrache:


cdjackson:
I wonder if we shouldn't combine this with the current "read" command? Currently the read command can take a list of cluster IDs - maybe we introduce * here to read all? What do you think? I think it's nicer to avoid too many extra commands if we can...

mikomarrache:
I've modified the existing read command and removed the new readall command.

cdjackson:
Thanks.

PR #1171, opened 2020-11-21
channel default UNKNOWN so that quietest channel is selected

mikomarrache:
Currently, using the console application, we could trigger selection of the quietest channel by specifying channel -1. However, I think it would be more intuitive to not provide channel argument at all so that quietest channel is selected.


cdjackson:
Agreed - makes sense - thanks.

Issue #982, opened 2020-02-05
Possible NPEs in EmberNcp


triller-telekom:
One example is:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/7533a7de7d26eb81c3f79a31d43bf4fa797a4e44/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/EmberNcp.java#L375-L381

According to `EzspSingleResponseTransaction` the `response` might not have been initialized and thus `transaction.getResponse()` might return `null`.

I do not know if this is an edge case where the device doesn't fully support reading out the key but for many responses in the `EmberNcp` class it is assumed that it is not `null`. Otherwise it will run into NPEs while calling `toString()` or `getStatus()`.

However we experienced such an edge case, hence I am opening this issue for discussion.

cdjackson:
I think most of these should not result in NPE, but you are right that there is the potential here. I think I've fixed a few in the past, and my suggestion is to fix them as/if they turn up - although if you fancy going through and fixing them all, then that's also fine :)

PR #1176, opened 2020-12-02
Fix NPE in EmberNCP::setConfiguration

dschall:
This change fixes a NullPointerException in EmberNCP::setConfiguration when the response is null.
See #982.


cdjackson:
#982

PR #1191, opened 2021-01-02
Clarified duration method parameter in ZigBeeNetworkManager.java

Ziver:
The duration input for permitJoin(...) method was not clear on the time value


cdjackson:
Thanks.

PR #1192, opened 2021-01-02
JSerialComm implementation of serial port

Ziver:
I've had issues with the included serial port library where I get a JVM crash on startup on my Windows 10 computer.
So I implemented a Serial port based on the jSerialComm library which is a still actively maintained and works on latest windows 10 computers.

I have tested the code on a cc2531 dongle but not on the other devices as they are not available to me. 

I'm not sure about the class name "ZigBeeJSerialCommPort", let me know if there are any better suggestions.

### DIFF
--- a/com.zsmartsystems.zigbee.serial/pom.xml
+++ b/com.zsmartsystems.zigbee.serial/pom.xml
@@ -23,11 +23,11 @@
             <artifactId>com.zsmartsystems.zigbee</artifactId>
             <version>1.3.11-SNAPSHOT</version>
         </dependency>
-        
+
         <dependency>
-            <groupId>org.scream3r</groupId>
-            <artifactId>jssc</artifactId>
-            <version>2.8.0</version>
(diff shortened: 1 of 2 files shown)


cdjackson:
Sorry - I've not processed this...

My view on this is we do not need to have multiple serial implementations within the library since this is generally not used here (only in the demo application). Now, I'm certainly not against changing the current implementation to a different library - serial libraries in Java are generally terribly supported, so if this updated version is being supported, that would be great and we could consider changing if it makes sense, but I don't think we need multiple implementations.

Ziver:
Thanks for the reply, I have updated the code based on your feedback and changed the default serial implementation to use jSerialComm.

I have had very god experience with the jSerialComm library where it supports many platforms and is still being actively supported. https://github.com/Fazecast/jSerialComm

Btw, I removed the SerialPortEventListener interface as I did not see any usage of it, let me know if this interface is still required.

PR #1165, opened 2020-11-19
Add ability to request NCP source route table

mikomarrache:

PR #1195, opened 2021-01-11
Add Ember routing information commands

cdjackson:
@mikomarrache I didn't hear back from you in #1165, so I propose this PR as a replacement. I've not fully tested this, and it's quite possible the dumping of the source route table is not correct - I only have a limited setup here at the moment due to moving, but welcome your thoughts.


mikomarrache:
Hi, sorry for the late reply...

I'm fine with it. I guess you can merge to master and we will test. Thanks!

cdjackson:
Ok, thanks @mikomarrache

Issue #1157, opened 2020-11-08
Zigbee Console netstart command is broken


tunjid:
**Describe the bug**

The `netstart` console command always errors out.

Line 622 of the ZigeeNetworkManager contains

```java
        if (networkState != ZigBeeNetworkState.INITIALISING) {
            logger.error("ZigBeeNetworkManager startup: Can't be called when networkState={}", networkState);
            return ZigBeeStatus.INVALID_STATE;
        }
```
Since at the time the console is running the network is always online, netstart will always fail.

**To Reproduce**B
Attempt to use the `netstart` console command

**Expected behavior**
The `netstart` command is available.

**Additional context**
Seems related to #891
I can submit a PR for this, I believe a check for wether it is being reinitialized should suffice?

cdjackson:
There is a command called `netreinitialise` (or something close to this) that you must run first. This sets the states correctly.

tunjid:
I'm not sure what it is, but when I use the `reinitialize` command before the `netstart` command, ZigBee routers are never able to join the network. The same happens if I call `networkManager.initalize()` directly.

If  don't reinitialize the network and just call `netstart` directly having commented the above out,  routers can join the network.

Are there things that need to be reconfigured after a `reinitialize`? The source seems to just update a flag when `initialize`. is called.

cdjackson:
I'm not really sure what you are doing, so it's hard to know how to help.

Normally, you should not need to use the netstart command - just initialise the system through the command line args.  The netstart command was added for some specific use case when we were testing a system as a router, and needed to connect to different networks.

It sounds from your description that you are creating a network, so I would simply recommend to use the command line arguments to initialise the coordinator.

tunjid:
I was able to work around it it by invoking the methods on the ZigBeeTransport(dongle) directly. Pardon the Kotlin:

```kotlin
        networkManager.zigBeeTransport.apply {
            updateTransportConfig(TransportConfig().apply {
                addOption(TransportConfigOption.DEVICE_TYPE, DeviceType.COORDINATOR)
            })
            out.println("ZigBees status is: ${startup(true)}")
        }
```

Thanks so much for your help!

cdjackson:
Really this shouldn't be needed. If you have a suggestion for a change in the console, then please feel free to provide an update, but you should be able to initialise a network now through the console cli - this is how the system is designed to work and the netstart function is really only there for "special occasions".

tunjid:
[...]

PR #1211, opened 2021-03-23
Set default argument for join enable to 254

tunjid:
> As per ZigBee 3, a value of 255 is not permitted and will be ignored

Addresses https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/1157#issuecomment-805187856


cdjackson:
Thanks.

PR #1214, opened 2021-04-01
Fix network backup command (console)

mikomarrache:
I also had to remove a condition check in ZigBeeNetworkManager that was preventing the restore.


mikomarrache:
I see that removing the condition caused a unit test to fail.

What do you think of removing the test?

cdjackson:
I would prefer not to remove this condition "just" to allow the backup to work. Maybe we should take a look at the restore workflow to understand if that can be improved or if there is a better way to approach it.

I guess you remove this as you want to call `startup` without `initialise`? Why?

mikomarrache:
We simply use the existing network restore command, which doesn't call `initialize`.

Also, the `initialize `method includes the following check at the beginning:

```
synchronized (this) {
    if (networkState != ZigBeeNetworkState.UNINITIALISED) {
        return ZigBeeStatus.INVALID_STATE;
    }
    setNetworkState(ZigBeeNetworkState.INITIALISING);
}
```

In my case, the network manager is already initialized so that won't work.

sraillard:
Without removing this test, the netbackup command (in restore mode) of the console can't work because when the console application is launched, the Zigbee network is initialized. 

Or, adding a command line option to skip the network initialization when starting the console application, so the initlization can be done later using the netbackup command?

PR #1216, opened 2021-04-01
Ember Dongle Firmware Update: read InputStream up to its end

### LINKED ISSUE #1215: EmberFirmwareUpdateHandler should read InputStream to its end
Currently the `EmberFirmwareUpdateHandler` receives an `InputStream` and expects to be able to read a multiple of 128bytes plus some rest from it.

Depending on the implementation of the `InputStream` it may act as a proxy to some remote server and implement a different chunk size.

At the moment, we treat a "non full" chunk, i.e. one that is lower than the expected 128 byte as "the end":

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/114454bbcea7f56a43b8a3c1dd59d44315db5cef/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/EmberFirmwareUpdateHandler.java#L274

IMHO we should always read to the end of the stream (`read()` returns `-1`) until we treat the stream as "finished".

triller-telekom:
Fixes #1215


cdjackson:
Thanks @triller-telekom, LGTM - long time no see - I hope you're doing well...

PR #1220, opened 2021-04-13
ZclOtaUpgradeServer.getCurrentFileVersion() should not use caching

### LINKED ISSUE #1219: ZclOtaUpgradeServer.getCurrentFileVersion() should not use caching
In the Javadoc the comment clearly says that there will be no caching involved and one will always get the latest value.

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/91082224e645301292043b51dec8885f3eae7c46/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java#L522-L533

However, the `readValue(Long.MAX_VALUE);` part makes it return whatever value that is currently stored for that attribute.

One use case where this does matter is openHAB: https://github.com/openhab/org.openhab.binding.zigbee/blob/7efcd7a26c86625d48a7f8978ebcd4a082b3a0af/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java#L1040

After a device firmware update, it requests the latest version. If one does a downgrade/upgrade of a firmware from one that supports the current version attribute, to one that does not support it, we will end up using the version number from the previous firmware and not the current (unknown) one.

In the scenario above we would have used this branch in the `completeUpgrade()` method:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/91082224e645301292043b51dec8885f3eae7c46/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java#L497-L503

So, if I understand this correctly, we should replace the `readValue(Long.MAX_VALUE);` with `readValue(0);`.

triller-telekom:
Fixes #1219

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/otaserver/ZclOtaUpgradeServer.java
@@ -529,7 +529,7 @@ public class ZclOtaUpgradeServer implements ZigBeeApplication, ZclCommandListene
      * @return the current firmware version on the remote device
      */
     public Integer getCurrentFileVersion() {
-        return (Integer) cluster.getAttribute(ZclOtaUpgradeCluster.ATTR_CURRENTFILEVERSION).readValue(Long.MAX_VALUE);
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks @triller-telekom

PR #1229, opened 2021-05-07
IAS client extension - handle case where node leaves the network

mikomarrache:
We made a simple test:

- an IAS device joins the network
- the enrollment is performed correctly (the _nodeAdded_ callback is called and the application is registered)
- after a while, the device is factory reset
- at some point, the device joins the network again but not IAS enrollment is performed

First, we saw that the _nodeAdded_ callback wasn't called in this case. I guess this is related to the fact the node is still in memory but its state set to OFFLINE. Therefore, this PR proposes to also implement the _nodeUpdated_ callback.

Second, even after the _nodeUpdated_ callback is implemented, enrollment is still not performed. This is due to the fact the _ZclIasZoneClient_ application is already registered for the IAS cluster on the specific endpoint. Therefore, this PR proposes to implement the _ZigBeeAnnounceListener.deviceStatusUpdate_ callback so that we can remove the application from the endpoint where the device status becomes _DEVICE_LEFT_.


cdjackson:
I'm a little worried about the consequences of this change. IIRC, when a child device (SED) changes parent, you will get a DEVICE_LEFT notification, quickly followed by a new DeviceAnnounce. However the device will be asleep again, and now while nothing has changed (other of course than the child has a new parent), you will now require to perform the IAS enrolment again. This will fail - the device might still work ok since it is really still enrolled, but the library will not be able to communicate and this is likely to choke communications somewhat.

So, I think here you solve one potential problem, but cause another - I guess the question might then be which is worse.

sraillard:
The parent router for the SED may send a DEVICE_LEFT notification if it think the device has left the network, so I think the best to remove the application associated to the endpoint is to catch the UNSECURE_JOIN event instead of DEVICE_LEFT.

The UNSECURE_JOIN event is sent when a new association is starting and the device should be without any IAS enrollment (like a factory reset). This will allow the IAS configuration to be done again.

Issue #690, opened 2019-07-21
RFC: Modify ZigBeeNodeListener


cdjackson:
Currently the ```ZigBeeNodeListener``` interface contains method for ```nodeAdded```, ```nodeUpdated``` and ```nodeRemoved```. It has been proposed in #649 to add a new ```nodeDiscovered``` method to be called when node discovery information changes.

I'm a little wary of adding new methods each time we want to notify of something new, and instead propose to change the interface to have a single method, and an enumeration stating the reason for the notification.

eg -:

```
nodeListener(ZigBeeNodeListenerNotification notification, ZigBeeNode node);

ZigBeeNodeListenerNotification {
    /*
     Node has been added to the network
     */
    ADDED,

    /*
     Node discovery data has been updated. This includes new endpoints or clusters being added to the node.
     */
    DISCOVERED,

    /*
     Node has been updated. This includes any changes to information such as routes and neighbours.
     */
    UPDATED,

    /*
     Node has been removed from the network
     */
    REMOVED
}
```

I do welcome comments on this approach versus the option of simply adding new methods.

triller-telekom:
Sorry for the late reply.

To me this sounds like a reasonable approach. Plus point of this approach is that we have to break the API now once and afterwards we can simply extend this enumeration to introduce further reasons for notifications.

mikomarrache:
Not having to rely on enums is generally a better approach. Enums introduce all sort of problems (client code may rely on ordinals, enum names, etc) while we don't get such issues using methods. For example, if I store the events triggered by this listener in a database, I would use the ordinal or the enum name. If enum order or entries names change, that would break client code (though client code still compiles).

Also, adding a method to an interface doesn't break the API as long as the method is declared as default (in this case with a no-op implementation).

I would go with the new method approach.

cdjackson:
We already rely on enums in other listeners such as the network state listener, and there is quite extensive use of enums throughout the framework. If we never used enums for notifications, we'd have a lot of methods so the problem is it's not scalable IMHO.

If a user is importing the framework, then the enum is available, and it's generally not recommended to use the ordinal on enums in any case as this can always change - again this would be the same if you use the current enums.

mikomarrache:
> We already rely on enums in other listeners such as the network state listener, and there is quite extensive use of enums throughout the framework. If we never used enums for notifications, we'd have a lot of methods so the problem is it's not scalable IMHO.

[...]

PR #1231, opened 2021-05-07
ZigBeeNetworkManager.refreshnode small fixes

mikomarrache:


mikomarrache:
There is one test failing, but I'm not able to make it fail locally so I'm not sure what is wrong here.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java
```diff
@@ -1767,24 +1766,23 @@ public void run() {
             return null;
         }
 
-        if (node.getNodeDescriptor() != null && networkNodes.get(node.getIeeeAddress()) != null && Objects
-                .equals(networkNodes.get(node.getIeeeAddress()).getNodeDescriptor(), node.getNodeDescriptor())) {
+        if (currentNode.getNodeDescriptor() != null) {
```
cdjackson, line 1769: This change will mean that the transport layer `setNodeDescriptor()` method gets called every time `refreshNode`  is called. This check should at least check that `node.getNodeDescriptor()` is not null - ie the node descriptor is being updated by the call to `refreshNode`.

cdjackson:
As a matter of interest, why did you change this? There is no real change to the code - you just use the `currentNode` instead of `node` to print the IEEE address, but given that these are guaranteed to be equal, this is the same thing. I don't mind changing it - maybe it's better is some view, but I'm wondering if it caused a bug or you just changed it for some other reason?

The only material change is to the comparison of the `NodeDescriptor` which is now different. I think the old check is more accurate as it checks if the node descriptor has changed - now it will falsely send updates as it only checks if it's not null.

mikomarrache:
No. We started discussing internally about the meaning of the nodeAdded, nodeUpdated and nodeRemoved callbacks. Then we saw that nodeAdded was called multiple times which makes this callback meaning not so clear - i.e we expect this callback to be called only once since a node is added only once to a network. I proposed to remove one call in PR #1230. But there is still another call which is made after discovery is performed. I still don't understand why this call should be made after discovery?

cdjackson:
> Then we saw that nodeAdded was called multiple times

This is historical and I think it is documented. IIRC the calls will be made when the device is first discovered, and then again when the discovery of the descriptors etc is complete. To avoid breaking existing users code, we had to stick with the two calls, but at this stage I don't mind changing this. There is also #690 which proposes to change the notification to a single call to try and consolidate this so it might be good to do this at the same time.

> I still don't understand why this call should be made after discovery?

As above - the two calls were introduced to avoid breaking existing users at the time.

mikomarrache:
So, if I understand correctly, #690 proposes to use a separate callback (nodeDiscovered - method or enum) to notify that node discovery has been performed. Again, it relates to the discussion about where discovery should be done (extension or core). As I said in other tickets, I don't understand why the network manager is aware about discovery. I'm also not sure why a nodeDiscovered callback is required. Client code could use nodeUpdated and take decisions based on that.

Maybe what is missing here is passing old node state in the nodeUpdated callback. I think that would be a more generic approach.

cdjackson:
> So, if I understand correctly, #690 proposes to use a separate callback (nodeDiscovered - method or enum) to notify that node discovery has been performed. 

It proposes to use a single method to be called when there is a node update and how the nodeAdded, nodeUpdated and nodeRemoved callbacks are provided. It is not about adding a new method relating to discovery. IIRC is proposes to use a single callback method, and a reason, rather than multiple methods and splits the dual "added" calls into two phases - once when the node is first heard, and once when the node information is known.

> Again, it relates to the discussion about where discovery should be done (extension or core).

Not really. The node is updated not only when it is discovered. #690 is simply about changing the way that node updates are notified - it's not related in any way to discovery (other than of course discovery is one of the many reasons that a node listener will be called).

> As I said in other tickets, I don't understand why the network manager is aware about discovery.

Fair enough, but I think I tried to explain this, and it's not really related to this issue. I've said that I'm happy to discuss discover improvements, but I still think it should be a separate "addon" as some people do not want this, and others may want to do something different. In any case, it is not related to this issue.

> Maybe what is missing here is passing old node state in the nodeUpdated callback.

I need to think about how this is used. While at first thought, it seems like it might be a good idea, I'm wondering if it is really useful. To use it, you would need to perform a comparison to see what has changed - which is what is already done when the node is updated.  How do you see passing the original and new node structure helping?

cdjackson:
I got distracted by the conversation which is not really related to this PR, and the changes here are fine so merging.

PR #1233, opened 2021-05-07
Rename ZigBeeNetworkManager.updateNode method to better express its functions

mikomarrache:
Breaking changes. I think that would be okay for v1.4.


cdjackson:
Let me think about this first, but I agree it is probably a good idea.

PR #1235, opened 2021-05-07
Move rediscover logic outside network manager

mikomarrache:


cdjackson:
Why do you want to move this? This will be a breaking change, and I'm pretty sure that it will upset a few users.

mikomarrache:
This function makes the network manager aware of the discovery extension. Why is it the responsibility of the network manager to expose the rediscover function? It is part of a set of changes I would like to propose around the relation between the "core" and the discovery extension.

cdjackson:
I'd suggest to discuss this further. As I said, removing this method will break peoples code as you are removing functionality.

As I said elsewhere - I'm not against changes to discovery - I feel it does need improvement and there are some open issues already in this area. I would suggest to discuss these changes in an issue before implementing them though.

PR #1243, opened 2021-06-08
Refactor Ember ASH receive thread

cdjackson:
This PR adds a new thread for handling the ASH communications with the Ember chip. This thread handles the ASH protocol, and places any received packets into a queue which can then be processed in a separate thread.

The ASH thread has a high priority to try and ensure it processes communications with the NCP in a timely manner. We have seen some communication delays on some slower systems that can cause the NCP to throw an exception and this is an attempt to resolve this by ensuring that the NCP communications is serviced at a high priority.

I will leave this a couple of days before to allow others to comment.


cdjackson:
I've given this a reasonable amount of testing, and had reports from the organisation that reported this issue that it has helped to eliminate NCP asserts, so I'll merge this now.

PR #1260, opened 2021-09-09
Fix ZigBeeDongleTelegesis to make FW update start

### LINKED ISSUE #1259: Firmware update for Telegesis dongle doesn't start
The FW update for the Telegesis dongle fails with a InvocationTargetException when sending the TelegesisBootloadCommand. 

```
java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.FutureTask@1b1051a rejected from java.util.concurrent.ThreadPoolExecutor@1adda31[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 22]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830)
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:134)
	at com.zsmartsystems.zigbee.dongle.telegesis.internal.TelegesisFrameHandler.sendRequestAsync(TelegesisFrameHandler.java:542)
	at com.zsmartsystems.zigbee.dongle.telegesis.internal.TelegesisFrameHandler.sendRequest(TelegesisFrameHandler.java:553)
	at com.zsmartsystems.zigbee.dongle.telegesis.ZigBeeDongleTelegesis.updateFirmware(ZigBeeDongleTelegesis.java:995)
```

Why it happens: In the ZigBee binding, TelegesisHandler.updateFirmware(...) calls zigbeeTransport.shutdown() before starting the firmware update. This method closes the TelegesisFrameHandler, which is then used in ZigBeeDongleTelegesis.updateFirmware to send commands. Consequently, it fails there because TelegesisFrameHandler is now using a terminated executor.

I compared the firmware update in ZigBeeDongleTelegesis and ZigBeeDongleEzsp and saw some differences:

ZigBeeDongleEzsp:
- In the method shutdown() the frameHandler is set to null.
- The method updateFirmware(...) expects the frameHandler to be null and creates a new instance in initialiseEzspProtocol, where some initialization stuff is also done.

ZigBeeDongleTelegesis:
- In the method shutdown() the frameHandler is not set to null.
- The method updateFirmware(...) expects the frameHandler to be not null and there is no reinitialization. And so it uses the existing instance that was previously closed.

I thought about aligning the Telegesis solution with Ember, which means
- ZigBeeDongleTelegesis.shutdown will set frameHandler to null
- ZigBeeDongleTelegesis.updateFirmware will expect the frameHandler to be null and then creates a new instance of the frameHandler.

I will create a PR with a suggested solution. I'm happy to hear your opinion or other ideas.

TomTravaglino:
This PR fixes a bug wich prevents the FW update for Telegesis dongle from being started. The bug occurs when using the zigbee binding where the frame handler is closed before starting the FW update. In ZigBeeDongleTelegesis.updateFirmware() the frame handler is then used to send commands where it fails as it's using a terminated Executor. This PR changes ZigBeeDongleTelegesis.updateFirmware() to reinitialize the frame handler before starting the update.

Signed-off-by: Tommaso Travaglino <tommaso.travaglino@telekom.de>

Resolves #1259

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/ZigBeeDongleTelegesis.java
+++ b/com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/ZigBeeDongleTelegesis.java
@@ -387,9 +387,7 @@ public class ZigBeeDongleTelegesis
         }
 
         // Create and start the frame handler
-        frameHandler = new TelegesisFrameHandler(this);
-        frameHandler.start(serialPort);
-        frameHandler.addEventListener(this);
+        initializeFrameHandler(serialPort);
 
         initialiseCriticalConfiguration();
 
@@ -429,6 +427,13 @@ public class ZigBeeDongleTelegesis
         }
 
         return ZigBeeStatus.SUCCESS;
+    }
+
+    private void initializeFrameHandler(ZigBeePort serialPort) {
+        // Create and start the frame handler
+        frameHandler = new TelegesisFrameHandler(this);
+        frameHandler.start(serialPort);
+        frameHandler.addEventListener(this);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks @TomTravaglino LGTM

PR #1269, opened 2021-10-13
Explicitly disallow rejoins using a well-known key

### LINKED ISSUE #1268: Is the Default Ember key exchange policy change a security issue?
Doing some security analysis of how we use ZigBee in our product made me stumble across a change of the default policy for network rejoins:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/962/files

We had set this flag in the same way for our product due to some issues with I think Osram ZigBee 3 bulbs and I had left a comment on it:
```
// We have some ZigBee 3 bulbs that ask our coordinator which stackCompliance level it supports. Apparently our
// Ember dongles support ZigBee 3 features and thus the devices decide to request a trust center link
// key from them. Therefore we change the policy to allow sending out the link key to the device.
```

But now as I read the Silabs documentation about it I am worrying whether this setting is actually a good idea in the home automation world, were all (old) devices use the well-known Alliance09 link key...

The Silabs documentation https://www.silabs.com/documents/public/application-notes/an1233-zigbee-security.pdf on page 13 says:

> The recommendation to silently reject the Trust Center Rejoin as opposed to explicitly reject the Trust Center rejoin is necessary to maintain compatibility with pre-Zigbee 3.0 devices. Pre-Zigbee 3.0 devices may try Trust Center Rejoins without considering whether an updated link key has been obtained previously. If an explicit rejection is done, the device may drop off the network entirely while a silent rejection would allow the device to try a Secure Rejoin later and succeed.
>
> The Trust Center’s behavior will ultimately determine the security of the Trust Center Rejoin attempt. This can be controlled via the emberTrustCenterJoinHandler() callback on the SOC or via the ezspSetPolicy() API configuring the EZSP_TRUST_CENTER_POLICY for an NCP. To silently ignore the Trust Center Rejoin on an SOC device, the emberTrustCenterJoinHandler() must return EMBER_NO_ACTION. To silently ignore the Trust Center Rejoin on an NCP, the host must set the EZSP_TRUST_CENTER_POLICY to EZSP_IGNORE_TRUST_CENTER_REJOINS.

Before #962 the `EzspPolicyId.EZSP_TC_KEY_REQUEST_POLICY` was set to `EzspDecisionId.EZSP_DENY_TC_KEY_REQUESTS` which probably leads to the issue stated by the silabs documentation, i.e. `If an explicit rejection is done, the device may drop off the network entirely`.

Setting it to `EzspDecisionId.EZSP_ALLOW_TC_KEY_REQUESTS_AND_SEND_CURRENT_KEY` as the code is doing now means that we send the network key to any rejoining device if it presents us a valid link key (i.e. the Alliance09 one for example), right?

Shouldn't we instead set the value to `EzspPolicyId.EZSP_IGNORE_TRUST_CENTER_REJOINS` as stated in the documentation? As I understand from it, this would force the devices to try a secure join, i.e. they need to know the current network key before they are allowed to rejoin the network

triller-telekom:
Fixes #1268

### DIFF
--- a/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java
+++ b/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/ZigBeeDongleEzsp.java
@@ -325,6 +325,8 @@ public class ZigBeeDongleEzsp implements ZigBeeTransportTransmit, ZigBeeTranspor
         stackPolicies.put(EzspPolicyId.EZSP_TC_KEY_REQUEST_POLICY,
                 EzspDecisionId.EZSP_ALLOW_TC_KEY_REQUESTS_AND_SEND_CURRENT_KEY);
         stackPolicies.put(EzspPolicyId.EZSP_TRUST_CENTER_POLICY, EzspDecisionId.EZSP_ALLOW_PRECONFIGURED_KEY_JOINS);
+        // this effectively sets the rejoin with a well-known key to 0 = OFF
+        stackPolicies.put(EzspPolicyId.EZSP_TC_REJOINS_USING_WELL_KNOWN_KEY_POLICY, EzspDecisionId.EZSP_ALLOW_JOINS);
         stackPolicies.put(EzspPolicyId.EZSP_MESSAGE_CONTENTS_IN_CALLBACK_POLICY,
                 EzspDecisionId.EZSP_MESSAGE_TAG_ONLY_IN_CALLBACK);
         stackPolicies.put(EzspPolicyId.EZSP_APP_KEY_REQUEST_POLICY, EzspDecisionId.EZSP_DENY_APP_KEY_REQUESTS);


triller-telekom:
`ZigBeeDongleEzsp` does have a `updateDefaultPolicy` method where an application can set anything, like I mentioned in my comment https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/1268#issuecomment-942076199

So I don't see any reason of not merging this.

cdjackson:
This feels inconsistent to me. We provide methods to change other configuration in this area (ie joining / rejoining) through the `TransportConfigOption`. This provides a consistent way to configure the dongle from the framework perspective and allows the configuration and policies to be changed on the fly. 

Also, the `updateDefaultPolicy` method only updates the policy in the default configuration - so this can only be configured prior to startup and doesn't allow someone (for example) to set a high security level, then reduce the security to make a some network change, and then increase the security again. It could only be done by restarting the system each time.

cdjackson:
I have checked the default on this policy, and the default value is set to `00` so this will not be a breaking change and I'm therefore happy to merge.

Issue #1242, opened 2021-06-06
CC2531 always returns transaction ID 0


Ziver:
I'm using this project in a custom project with the CC2531 dongle, I'm able to pair devices but no descriptors or endpoints where not showing up including for the dongle itsself.

From my investigation it looks like the root cause of the issue is related to the pull request https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1128 where ZigBeeTransaction.java changed from:
`
if (responseMatcher.isTransactionMatch(command, receivedCommand)) {
`
To also verify the transaction ID:
`
if (Objects.equals(command.getTransactionId(), receivedCommand.getTransactionId())  && responseMatcher.isTransactionMatch(command, receivedCommand)) {
`
This change makes total sense but it looks like there is some issue with the CC2531 dongle where it always returns transaction ID as 0 for some responses which causes the transaction to not complete and instead timeout and fail.

**Some logs:**
Non working 1.3.11 version: [1.3.11_not_working.log](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/files/6603495/1.3.11_not_working.log)
Where the transaction fails (TID=04):

```
2021-06-06 03:41:05.059 [FINE]    ZigBeeNetworkManager        : RX CMD: NodeDescriptorResponse [0000/0 -> 0000/0, cluster=8002, TID=00, status=SUCCESS, nwkAddrOfInterest=0000, nodeDescriptor=NodeDescriptor [apsFlags=0, bufferSize=80, complexDescriptorAvailable=false, manufacturerCode=0000, logicalType=COORDINATOR, serverCapabilities=[PRIMARY_TRUST_CENTER], incomingTransferSize=160, outgoingTransferSize=160, userDescriptorAvailable=false, frequencyBands=[FREQ_2400_MHZ], macCapabilities=[ALTERNATIVE_PAN, FULL_FUNCTION_DEVICE, MAINS_POWER, RECEIVER_ON_WHEN_IDLE], extendedEndpointListAvailable=false, extendedSimpleDescriptorListAvailable=false, stackCompliance=21]]
[...]

PR #1272, opened 2021-11-09
Fix for CC2531 missing transaction ID

Ziver:
The commit adds a map to store sent transaction IDs that are later injected into received payloads when CC2531 controller is used.

This PR should fix issue #1242


cdjackson:
Thanks. I can't easily test this so I assume it's all good. Can you please just sign the commit and I'll merge.

Ziver:
Commit should be signed now.

PR #1273, opened 2021-11-09
Switched default serial implementation to JSerialComm

Ziver:
My previous PR was not compiling after some updates in master so this is a rebase. Feel free to close this if this is not wanted change in the console/test application.

The aim of this PR is to move over to a more regularly updated and maintained serial library as the previous lib has stopped receiving updates since 2014. I'm personally getting JVM crashes when trying to run the console application on Win 10/Java 11

I've had very god experience with the jSerialComm library for many years where it supports many platforms and is still being actively updated. https://github.com/Fazecast/jSerialComm

Also changed baud rate input parameter to be mandatory in console application as it crashes without defining it.


cdjackson:
You are of course free to use whatever library you like in your application - the library used in this repository is only used in the example console application and doesn't dictate anything for users of the library. The library ensures an abstraction with the use of the `ZigBeePort` interface.

I doubt I will incorporate PR this as the API is a bit less standard. I might migrate to using PureJavaSerial at some point, but also need to ensure it can run at 2Mb/s as that is also required here.

PR #1311, opened 2022-03-01
Changed shutdown methods of drivers to always close handlers

### LINKED ISSUE #1305: EZSP ASH handler is not always closed.
If the ASH handler fails to initialise correctly, then it doesn't get closed when the Ember driver shuts down. A flag was introduced in #1287 to only perform the shutdown after it was configured and this is set at the end of the `initialize` method. If there is a failure  in the `initialize` method, then the flag isn't set, and then when `shutdown` is called, the `frameHandler` is not closed - leaving the serial port open. In the log we see a continual `ASH: Reconnect` being loged in the `AshFrameHandler`.

@TomTravaglino just pinging you here as I recall we had quite some discussion about this. I've not looked at this in detail yet, but I guess that `shutdown` needs to reverse `initialize` - so if the port is open, we should close it, and if there is a handler open, it should be closed...

TomTravaglino:
Resolves #1305 

This PR fixes the issue where `shutdown` doesn't close the handlers when the initialization was not successful.


cdjackson:
Thanks @TomTravaglino - much appreciated. I had not had time to look at this, but it looks like a good (and simple) approach.

PR #1331, opened 2022-06-08
Implement a longer initial timeout to allow for the stack to initialise

cdjackson:
@tejas-n please confirm this resolved the issue.


tejas-n:
Thanks @cdjackson. I just tried it out locally and it seems to fixing the problem.

PR #1333, opened 2022-06-10
Add test for ExtendedPanId.equals

lacinoire:
Hey 😊
I want to contribute the following test:

Test that `address.equals(new java.lang.Object())` is false when `equals` is called with the parameter `obj = new java.lang.Object()`.
This tests the method [`ExtendedPanId.equals`](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/0567bd1f6837d51adfd605dae7bcb7a776079c6f/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ExtendedPanId.java#L119).
This test is based on the test [`testConstructorString`](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/0567bd1f6837d51adfd605dae7bcb7a776079c6f/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/ExtendedPanIdTest.java#L44).

Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/ExtendedPanIdTest.java
+++ b/com.zsmartsystems.zigbee/src/test/java/com/zsmartsystems/zigbee/ExtendedPanIdTest.java
@@ -79,6 +79,12 @@ public class ExtendedPanIdTest {
     }
 
     @Test
+    public void testNotEquals() throws Exception {
+        ExtendedPanId address = new ExtendedPanId("17880100dc880b");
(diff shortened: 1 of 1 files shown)


cdjackson:
Hi. Thanks for this. I'm happy enough to merge this as you're right that notEquals is not tested at the moment.

PR #1346, opened 2022-09-12
Ezsp incoming network status handler (0xC4)

mikomarrache:


cdjackson:
Looking at the latest UG100, this frame (0x00c4) is unassigned. What reference did you use to define this?

mikomarrache:
Yes - this is an information we got via the support.

image

cdjackson:
Ok, thanks. I wonder if this really means it was removed in "2.6" or "6.2". Anyway, the library doesn't use this at the moment so it should generally be ok either way.

PR #1354, opened 2022-10-01
Fix failing macCapabilities test by ensuring macCapabilities is not null

cdjackson:


cdjackson:
@mikomarrache FYI there was a failing test and my solution here was to ensure that the set is not null. Hopefully this doesn't impact your system.

PR #1366, opened 2022-12-13
Fix ConcurrentModificationException while iterating ZigBeeTransactionQueue.queue

mikomarrache:
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/1365

PR #1369, opened 2022-12-31
Bump animal sniffer to check against Android level 21

cdjackson:
#1366 uses class `java.util.concurrent.ConcurrentLinkedDeque` which is only available from Android API level 21. This PR bumps animal sniffer to check against level 21.

@tejas-n what version of Android are you guys using these days?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -240,7 +240,7 @@
 			<plugin>
 				<groupId>org.codehaus.mojo</groupId>
 				<artifactId>animal-sniffer-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


cdjackson:
I'll merge this to fix the build. If it turns out that #1366 was a breaking change for some Android users we'll deal with that separately. I'm hopeful that most users have moved past this by now ;)

tejas-n:
Hi, @cdjackson. Our old devices which are still in production run Android API level 22 :)

PR #1371, opened 2023-01-01
Add ZigBeeUtcTime class and client/server

cdjackson:
This is something I wrote a while back but didn't create a PR so I need to try and remember how it all works :)

This creates a `ZigBeeUtcTime` class for managing time on a remote device. It then supports a `ZigBeeTimeExtension` which in turn can provide a local `ZclTimeClient` or `ZclTimeServer` to manage the time on the remote device.

The `ZclTimeServer` is relatively simple in that it just responds to requests from the device.

The `ZclTimeClient` is in itself also simple in that it provides methods to set the time on the remote.

The `ZigBeeTimeExtension` tries to automate the time management by working out drift rates, and setting the clock proactively if the time has drifted outside spec and the device didn't update the time through a request to the `ZclTimeServer`.

There is also a `ZigBeeConsoleTimeCommand` console command to get or set the time, and display the various information.

This is potentially a breaking change if people are using time at the moment since this changes some interfaces from `Calendar` to `ZigBeeUtcTime`.

I welcome comments on this and will wait a week or two before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,7 @@ Currently implemented extensions include -:
 - IAS CIE client (`ZigBeeIasCieExtension`)
 - OTA Upgrade Server (`ZigBeeOtaUpgradeExtension`)
 - Basic Cluster Server (`ZigBeeBasicServerExtension`)
+- Time management (`ZigBeeTimeExtension`)
 
 These provide basic functionality and can be extended as required to meet the application needs.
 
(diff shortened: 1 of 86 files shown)


dschall:
Hey chris, apologies, I still didn't have time to test.
Thanks for coding this up, I hope to be able to run some testing soon...

cdjackson:
@dschall I've updated this to use the `ZclCommandListener` - this should avoid multiple responses. I also updated some initialisation code to register the clusters and merged master into this so it's up to date.

PR #1380, opened 2023-03-23
Replace runnable wrappers by lambda expressions/method references

ViToni:
By switching to Java8 (#408) `Runnable` wrappers can be replaced be Lambda expression or even method references as done in this PR.

TODO:
 - `com.zsmartsystems.zigbee.console.main.ZigBeeConsole.TransactionQueuesCommand.process(ZigBeeApi, String[], PrintStream)` is using an empty `Runnable` to start a `Thread`. Wondering if this is just a CnP error or if it has more to it.


ViToni:
TODO:
 - `com.zsmartsystems.zigbee.console.main.ZigBeeConsole.TransactionQueuesCommand.process(ZigBeeApi, String[], PrintStream)` is using an empty `Runnable` to start a `Thread`. Wondering if this is just a CnP error or if it has more to it.

ViToni:
@cdjackson Would you mind taking a look at this one?

ViToni:
@cdjackson Thanks for the feedback. Didn't know about restrictions regarding Android / Lamba expressions (but I have to admit I'm not deep into Android regarding redevelopment).
But indeed I remember that some early versions of Java 8 had some issues regarding Lambda expressions but AFAIK these have been fixed for some time now.

Regarding the first Android version supporting Java 8, from what I found, it seems to be Android Nougat from August 2016 or respective [Android Studio 3.0.0](https://developer.android.com/studio/releases/past-releases/as-3-0-0-release-notes) from October 2017. The final release of Nougat was [October 2019](https://android.googlesource.com/platform/build/+/android-7.1.2_r39).
The [Java 8 support page](https://developer.android.com/studio/write/java8-support) mentions only serialization of Lambdas not being supported, not sure if serialization has been used anywhere in project on Runnables.

Don't know if this infromation helps on the decision though...

cdjackson:
I found that animal sniffer throws an error with your recent addition of a lambda in the queue management PR, so i guess that answers the compatibility question ;)

PR #1388, opened 2023-04-21
Remove lambda function to fix animal sniffer error

cdjackson:
Revert recent code change due to animal sniffer failure.

@ViToni FYI


ViToni:
Thanks for the feedback @cdjackson.
Again something I learned, didn't know about animal sniffer, yet.

After checking the animal sniffer configuration I don't really know how things are matching.
On one hand the project uses Java 8 (which would match better to Android Nougat (level 24/25) or even Android Oero with minimal level of 26).
On the other hand there is PR #1369 which bumped Android level up to 21, is this the max. signature for the whole project, or is it supposed to be only for `com.zsmartsystems.zigbee`?

While running `mvn animal-sniffer:check` on the root level project of current `master` I encountered #1390.

cdjackson:
As mentioned elsewhere, some of the projects are not applicable to Android. In general it's only the main library and the dongles. Code generators and the console are not applicable to Android.

Note also that Android isn't officially Java, so I don't think you can assume that features added in certain Java versions is added in Android as there are differences.

PR #1394, opened 2023-04-25
Introduced Thread Pool Optimizations

### LINKED ISSUE #1392: Optimize thread pools to spare resource consumption
**Is your feature request related to a problem? Please describe.**
A leak has been encountered in which a scheduled thread pool comprises > 400.000 tasks and > 300.000 tasks were in cancelled state but still remained in the thread pool. After analyzing the heapdump, it is recommended to optimize the thread pools efficiently by introducing several thread pool executor flags.

**Describe the solution you'd like**
The following flags should be set to better optimize an executor:

* `corePoolSize` - to be set to a minimum number of threads which will be always available even when there are no tasks assigned to the thread pool (for example, `1` as core pool size would also work efficiently)
* `maxPoolSize` - to be set to the maximum number of threads to be created on demand when tasks need to be executed
* `keepAliveTime` - to be set to enable `coreThreadTimeOut` for terminating threads if no tasks arrive within this keep-alive time
* `removeOnCancelPolicy` - to be set for removing tasks from `ScheduledThreadPoolExecutor` to remove the allocated threads when they are cancelled

**Describe alternatives you've considered**
Usual way of creating threads out of a thread pool is not optimized for memory constrained devices, hence, there is no other alternative than configuring the aforementioned options for thread pools

**Additional context**
The changes are required to be introduced in [ZigBeeExecutors](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/master/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeExecutors.java)

amitjoy:
fixes #1392

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeExecutors.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeExecutors.java
@@ -7,11 +7,16 @@
  */
 package com.zsmartsystems.zigbee;
 
+import static java.util.concurrent.TimeUnit.SECONDS;
+
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


cdjackson:
@amitjoy for information, does the problem you reported in #1392 actually still exist - ie is this PR actually needed? I'm not necessarily against merging this, but it's really a backstop and I'm wondering if the other PRs that have been merged by @ViToni actually prevent the threads being created in the first place?

amitjoy:
@cdjackson Actually, this PR complements the changes @ViToni has already introduced. The important change @ViToni introduced is to enable removal of `ScheduledFuture` instances from the working queue that are already cancelled but the change I introduced is to ensure that there should not be any idle thread waiting for tasks for longer time and consume resources perpetually. 

For example, here https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/51d650319ac4d790a5ddad6c30cda261402209a7/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeExecutors.java#L57 

we created a thread pool having a core pool size of `n` number of threads and when the executor is created all the `n` threads will be created by the JVM directly even if there may be `m` no of tasks where `m < n`. So, the `n - m` no of threads are parked idle and they will never be removed unless we allow timeout to them. So, now with the changes I introduced, the `n - m` no of threads will keep on waiting for `x` no of seconds (`keepAlive` timeout) and if there is no task assigned to the thread pool, the idle core threads will automatically be removed from the thread pool, therefore, spare resources.

Since, this is a library that can be used in any environment, I strongly believe that such changes are necessary to ensure less resource utilization even if they create thread pool with large number of threads.

cdjackson:
Thanks.

PR #1395, opened 2023-04-25
Integrate Error Prone for static code analysis

ViToni:
[Error Prone](https://errorprone.info/) provides compile time static code analysis.
It requires at least JDK 11 since it makes use of some compiler hooks which are only available starting with JDK 11.
The setup allows usage of a JDK up to version 15 (version 16 and higher require addtional configuration).

This change does NOT change the default compilation source and target.
To avoid any interference between code analysis and production code, there is a special profile enabling code analysis.

This profile can either be selected explicitly:
```sh
mvn -P error-prone-check clean compile
```

or implicitly by injecting a JDK in the version range [11,16)
```sh
JAVA_HOME=/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home mvn clean compile
```


ViToni:
Ideally this check would be executed in a pipeline, but I'm not firm with GitHub actions and the code analysis would require configuration of another JDK.

ViToni:
@cdjackson Would you mind giving it a look.

cdjackson:
Thanks @ViToni - looks good.

PR #1398, opened 2023-05-27
Fix warnings for rawtypes

ViToni:
Some rawtype usages are induced by method signatures of external APIs (XStream) and are suppressed as the cannot be avoided. Other occurences were fixed.


cdjackson:
Thanks @ViToni

PR #1399, opened 2023-05-28
Fix missing line break when outputting in class generators

ViToni:
Fixes the method `outputWithLinebreak` which doesn't output line breaks in some cases.

### DIFF
--- a/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeBaseClassGenerator.java
+++ b/com.zsmartsystems.zigbee.autocode/src/main/java/com/zsmartsystems/zigbee/autocode/ZigBeeBaseClassGenerator.java
@@ -286,9 +286,8 @@ public abstract class ZigBeeBaseClassGenerator {
                 len += word.length();
             }
 
-            if (len != 2 + indent.length()) {
-                out.println();
-            }
+            // loop or skipping loop doesn't end line so we do
+            out.println();
         }
     }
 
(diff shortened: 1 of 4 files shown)


cdjackson:
I think it would be good to also run the code generators, so that any changes are picked up as well. Otherwise we'll get unrelated changes next time someone makes a change.

ViToni:
The change is to preserve actual formatting. Running the generators without this change, changes formatting in `com.zsmartsystems.zigbee.dongle.telegesis/src/main/java/com/zsmartsystems/zigbee/dongle/telegesis/internal/protocol/TelegesisDeviceType.java`.
The other generators were changed for consistency.

cdjackson:
Ok, I'm not sure I understand how this change doesn't change something in the code? You say that without this change the files will change - doesn't that mean there is a change in some files? Or did the previous PR make some changes that are reverted without this PR?

Sorry - just trying to understand...

ViToni:
If one would run the Telegesis generator as it is right now, the generator output would differ from what is currently checked in. The output would also no longer look like as expected.
It would mangle empty Javadocs in ˋTelegesisDeviceTypeˋ. One can check the diff which would result from running the Telegesis generator code on ˋmasterˋ.
The diff led me to look at the code in the generator and for me it looked like a bug and I saw the same pattern used in the other generators.
With this PR applied the output stays stable, which means for Telegesis the output will look like the checked in version, even on succeeding runs. For the other generators there is no change in the output with / without this change.
My assumption is that some (XML) data for Telegesis triggered this bug. The data is untouched but the code now handles the situation which triggered the bug.
I hope this explains it better than my first attempt.

cdjackson:
Thanks for the explanation. All good.

PR #1419, opened 2024-02-22
Fixed NPE

### LINKED ISSUE #1418: NPE in ZigBeeNetworkDiscoverer
The following exception occurred in our runtime and it seems that it should have never happened. I think it would be best to introduce the relevant preconditions to impose API's expectations. 

I will try to open a PR for this.

```java
java.lang.NullPointerException: null
    at com.zsmartsystems.zigbee.app.discovery.ZigBeeNetworkDiscoverer.getAssociatedNodes(getAssociatedNodes:388)
    at com.zsmartsystems.zigbee.app.discovery.ZigBeeNetworkDiscoverer.access$500(access$500:48)
    at com.zsmartsystems.zigbee.app.discovery.ZigBeeNetworkDiscoverer$3.run(run:324)
    at java.util.concurrent.Executors$RunnableAdapter.call(call:511)
    at java.util.concurrent.FutureTask.run(run:266)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(access$201:180)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(run:293)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(runWorker:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(run:624)
    at java.lang.Thread.run(run:750)
```

amitjoy:
Added NPE invariant
fixes #1418
Signed-off-by: Amit Kumar Mondal <admin@amitinside.com>

### DIFF
--- a/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
+++ b/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
@@ -7,12 +7,15 @@
  */
 package com.zsmartsystems.zigbee.app.discovery;
 
+import static java.util.Objects.requireNonNull;
+
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -59,7 +62,7 @@ public class ZigBeeNetworkDiscoverer implements ZigBeeCommandListener, ZigBeeAnn
     /**
      * The ZigBee network manager.
      */
-    private ZigBeeNetworkManager networkManager;
+    private final ZigBeeNetworkManager networkManager;
 
     /**
      * The minimum time before performing a requery
@@ -82,6 +85,7 @@ public class ZigBeeNetworkDiscoverer implements ZigBeeCommandListener, ZigBeeAnn
      * @param networkManager the {@link ZigBeeNetworkManager}
      */
     protected ZigBeeNetworkDiscoverer(final ZigBeeNetworkManager networkManager) {
+        requireNonNull(networkManager, "Network manager cannot be null");
         this.networkManager = networkManager;
     }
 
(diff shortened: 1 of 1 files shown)


com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
```diff
@@ -385,8 +389,14 @@ private boolean getAssociatedNodes(final int networkAddress) throws InterruptedE
             // Request extended response, start index for associated list is 0
             final IeeeAddressRequest ieeeAddressRequest = new IeeeAddressRequest(networkAddress, 1, startIndex);
             ieeeAddressRequest.setDestinationAddress(new ZigBeeEndpointAddress(networkAddress));
-            CommandResult response = networkManager.sendTransaction(ieeeAddressRequest, ieeeAddressRequest).get();
+            Future<CommandResult> resultFuture = networkManager.sendTransaction(ieeeAddressRequest, ieeeAddressRequest);
+            if (resultFuture == null) {
+                logger.debug("Transaction manager has been shutdown");
```
cdjackson, line 394: I'm not 100% sure that this is correct? Firstly, I don't think that a null return here necessarily means that the transaction manager is shut down - although that is one possibility. Secondly though, if the transaction manager is shutdown and that is the cause of the null return, then this same message is already logged in `queueTransaction` so I'm not sure we need to have this log entry.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
```diff
@@ -385,8 +389,14 @@ private boolean getAssociatedNodes(final int networkAddress) throws InterruptedE
...
+                return false;
+            }
+            CommandResult response = resultFuture.get();
             if (response.isError()) {
+                logger.debug("Transaction command has not been successful");
```
cdjackson, line 399: I haven't checked the logging elsewhere, but I feel that this log entry probably isn't needed as the failure should already be logged. Maybe I'm wrong, so if you think it's needed let's keep it, but I would be tempted in that case to make the log more useful and print the `response` as part of the log entry as it will say if it timed out or failed.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
amitjoy: Yeah, you are absolutely right. We can simply get rid of this log message.

com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/app/discovery/ZigBeeNetworkDiscoverer.java
amitjoy: Since the failure is already logged elsewhere, we can simply remove the logging here as well.

cdjackson:
Thanks @amitjoy

